PR #58, opened 2017-03-30
Mapped component package sharing improvment 

Llewellynvdm:
Added owner and key control to export & import of components, also added buy and package link building to global options.


ro-ot:
Sorry @Llewellynvdm I missed this, but the repeatable fields are not being exported and imported correctly.

Issue #59, opened 2017-03-31
Repeatable fields export and import


ro-ot:
Repeatable fields are not being exported and imported correctly. Their ids are not being updated, and therefore mismatching occurred every time the field id change.

Llewellynvdm:
ahh I missed that one. Will push a fix soon. Thanks!

PR #61, opened 2017-03-31
Stable release of component mapped export and import feature

Llewellynvdm:
Resolves gh-59 the repeatable field id problem. Also improved the export and import to insure snippets and linked admin views are re-mapped on import of the component. Also fixed the target and match field conditions in admin views to re-map the ids correctly.


admin/models/joomla_components.php
```diff
@@ -438,27 +444,36 @@ protected function setData(&$user, &$db, $table, $values, $key)
...
+					// load to global object
+					$this->smartExport[$table][$item->id] = $item;
+					// set the custom code ID's
+					$this->setCustomCodeIds($item, $table);
 					// actions to take if table is admin_view
 					if ('admin_view' === $table)
 					{
 						// add fields
 						$this->setData($user, $db, 'field', $item->addfields, 'field');
-						// add admin views
```
ro-ot: Why did you remove this, what if the linked view is another components view?

Llewellynvdm: That option is not working at this time, so my idea was to just use the actual views of the component... I could add it back. At first it caused a loop, but I fixed that.

admin/models/joomla_components.php
```diff
@@ -438,27 +444,36 @@ protected function setData(&$user, &$db, $table, $values, $key)
...
 					$this->smartExport[$table] = array();
 				}
 				foreach ($items as $nr => &$item)
 				{
+					// set the data per id only once
 					if (isset($this->smartExport[$table][$item->id]))
 					{
 						continue;
 					}
+					// load to global object
```
ro-ot: Smart move to prevent the loop :+1:

Issue #63, opened 2017-04-01
Repeatable fields get added twice during import of package a second time.


Llewellynvdm:
I just noticed that the repeatable fields are being added a second time since the XML is changed and so the local field is not found.

Issue #62, opened 2017-04-01
Full encryption of JCB packages


Llewellynvdm:
We will have to insure that we also encrypt the the files, images and folders in the JCB packages. This will insure complete security.

PR #64, opened 2017-04-01
Fixes to JCB package importer.

Llewellynvdm:
Resolves gh-63 to prevent repeatable field from being created a second time. Resolves gh-62 to insure full encryption of JCB packages.


ro-ot:
@Peeripapo have you had time to test this release?

Issue #72, opened 2017-04-25
Upgrade to Joomla 3.7.0 broke some of the views


Llewellynvdm:
I noticed whit the upgrade to Joomla 3.7.0 there are some small display issues with some of the views, like the options icon on the dashboard does not look correct and views with titles above the tabs and next to the tabs are also not displaying as expected.

Will have this fixed with the next release of JCB.

Issue #71, opened 2017-04-24
Admin view edit hung up


JohnnyMcWeed:
Since I updated to the actual version, I can do the following procedure and the component builder doesn't work anymore:

- Make a new admin view
- Save the admin view
- Open it again to edit

It then starts to load, but never finish. (Only for newly created admin views)
J: 3.6.5   CB: 2.4.3   PHP: 5.6.26

Console error tells this:
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at e.JRepeatable.a.$input.data.a.buildRows (repeatable.js:6)
    at e.JRepeatable.a.$input.data.a.init (repeatable.js:6)
    at new e.JRepeatable (repeatable.js:6)
    at HTMLInputElement.<anonymous> (repeatable.js:6)
    at Function.each (jquery.min.js:2)
    at a.fn.init.each (jquery.min.js:2)
    at a.fn.init.e.fn.JRepeatable (repeatable.js:6)
    at repeatable.js:6
    at dispatch (jquery.min.js:3)
    at r.handle (jquery.min.js:3)
```

Llewellynvdm:
I am not getting the same error on my side. Lets see.... hmmm

The **JRepeatable** function is a Joomla core function. I wonder if it will help to reinstall Joomla... not sure, but worth a try. Not a full reinstall, just to download the full Joomla 3.6.5 package and unzip it in your root folder and so override all files that are part of the core. Then simply delete the install folder, since no need to reinstall the DB again, this is not a DB error from what I can see in your console error.

Once you have done that, please download the latest [JCB 2.4.3](https://github.com/vdm-io/Joomla-Component-Builder/releases/download/v2.4.3/JCB_v2.4.3.zip) as well and reinstall it just for good measure. 

Now we are sure everything should be correct in terms of the core and the JCB component. Then do another test and let me know if that helped.

Honestly I have never seen this kind of error, and so I would say search your environment to insure that all is setup correctly. Check that all Joomla requirements are met. I suppose you have done that already, and that most of what I said my seem like an insult.

But please take my advice, I have often found that the smallest thing can cause major issues, and so double checking things is second nature in my DEV world...lol

I realize you said that things did work, until the upgrade, so if all the above remains fruitless. We have few options really, you can give me access to the server, and let me try and debug this. Second you try with a new fresh install of Joomla and JCB. You can also try to go back to the OLD version that worked.

[...]

PR #73, opened 2017-04-26
Few fixes to the compiler, and few bug fixes as metioned below

Llewellynvdm:
Resolves gh-71 bug, so that admin views that has no fields are still accessible. Fixed gh-72 to insure all components on Joomla version 3.7.0 works and looks correctly. Please just recompile your components for Joomla v3.7.0


admin/models/joomla_components.php
```diff
@@ -269,8 +292,17 @@ protected function smartExportBuilder()
 			$this->info['getKeyFrom']['website']		= $this->params->get('export_website', null);
 			$this->info['getKeyFrom']['license']		= $this->params->get('export_license', null);
 			$this->info['getKeyFrom']['copyright']		= $this->params->get('export_copyright', null);
-			$this->info['getKeyFrom']['buy_link']		= ($buy_link = $this->params->get('export_buy_link', null)) ? $buy_link . $this->params->get('export_buy_query', '&package=') . $this->packageName : null;
-			$this->info['getKeyFrom']['package_link']	= ($package_link = $this->params->get('export_package_link', null)) ? $package_link . $this->params->get('export_package_query', '&package=') . $this->packageName : null;
+			// making provision for future changes
```
ro-ot: What future changes? seems redundant to use an array if more then one components is exported. But interesting, looking forward to see where this will go. But yes this does give a better control on the buy link per component.

Issue #76, opened 2017-05-04
Error when trying to compile a component


nosfas:
Installed JCB on 2 different Joomla sites, one with Joomla 3.7 and the other with the 3.6.5 version.
Built the component as best as I could understand by looking at the demo on each joomla site.
Tryed to compile on both, neither completed.
Getting the following error on both Joomla compilations:
0 DateTime::__construct(): Failed to parse time string (-001-11-30T00:00:00+00:00) at position 7 (-): Double timezone specification
The Demo compilation works, but not anything else.

Managed to enable the Joomla debug options and got the call stack:
Call stack
#	Function	Location
1	DateTime->__construct()	JROOT/libraries/joomla/date/date.php:118
2	JDate->__construct()	JROOT/libraries/joomla/factory.php:530
3	JFactory::getDate()	JROOT/administrator/components/com_componentbuilder/helpers/compiler/b_Structure.php:799
4	Structure->getLastModifiedDate()	JROOT/administrator/components/com_componentbuilder/helpers/compiler/b_Structure.php:642
5	Structure->setDynamique()	JROOT/administrator/components/com_componentbuilder/helpers/compiler/b_Structure.php:342
6	Structure->__construct()	JROOT/administrator/components/com_componentbuilder/helpers/compiler/e_Interpretation.php:61
7	Interpretation->__construct()	JROOT/administrator/components/com_componentbuilder/helpers/compiler/f_Infusion.php:46
8	Infusion->__construct()	JROOT/administrator/components/com_componentbuilder/helpers/compiler.php:68
9	Compiler->__construct()	JROOT/administrator/components/com_componentbuilder/models/compiler.php:176
10	ComponentbuilderModelCompiler->builder()	JROOT/administrator/components/com_componentbuilder/controllers/compiler.php:79
11	ComponentbuilderControllerCompiler->compiler()	JROOT/libraries/legacy/controller/legacy.php:709
12	JControllerLegacy->execute()	JROOT/administrator/components/com_componentbuilder/componentbuilder.php:59
13	require_once()	JROOT/libraries/cms/component/helper.php:394
14	JComponentHelper::executeComponent()	JROOT/libraries/cms/component/helper.php:369
15	JComponentHelper::renderComponent()	JROOT/libraries/cms/application/administrator.php:98
16	JApplicationAdministrator->dispatch()	JROOT/libraries/cms/application/administrator.php:156
[...]

Issue #75, opened 2017-05-03
Dynamic Get on Joomla 3.7.0 issue


mwweb:
I discovered when attempting to finalize some gets on my component that the Selection box on Join View Tables, and also on Join DB Tables, isn't populating with the fields from the selected DB. (This is on the repeatable fields)

I did some testing, and the select box (Text Area) would populate with Joomla! 3.6.5, running JCB 2.4.3, but after upgrade to JCB 2.4.4 and Joomla 3.7.0 they no longer populate.

Even after a fresh install of Joomla 3.7.0 and JCB 2.4.4 there was no change.

Now, I am seeing the following in webmaster tools console:

jquery-migrate.js?eb66ae6…:23 JQMIGRATE: Migrate is installed with logging active, version 1.4.1
jquery-migrate.js?eb66ae6…:45 JQMIGRATE: jQuery.fn.load() is deprecated
migrateWarn @ jquery-migrate.js?eb66ae6…:45
jquery-migrate.js?eb66ae6…:47 console.trace
migrateWarn @ jquery-migrate.js?eb66ae6…:47

Note: I tested this on my localhost, and on separate hosting server with the same issues.

Llewellynvdm:
I will have a look at this, yet I did not have this issue as far as I know.

ro-ot:
Yes I also found this, must be that Joomla changed some of their conventions in the field naming.

PR #77, opened 2017-05-04
Joomla 3.7 fixes

Llewellynvdm:
Fixed gh-75 to insure the column names are loaded as before. Fix gh-76 by adding more checks to insure that the correct date is used in the getLastModifiedDate method.


mwweb:
I just tested for you on a test install, and the Dynamic Get fields are populating again.

Issue #100, opened 2017-06-11
Field Length Issue


mwweb:
I was testing adding my update server to my component, and was finding that the field wasn't letting me type everything. In the database, I could continue adding. What I discovered is the field in #__componentbuilder_joomla_component, update_server_ftp, and sales_server_ftp are set to text, but the form fields, in JCB -> Joomla Components -> Dynamic Integration, are set to maxlength="150", which appears to be what is "cutting off" input.

Issue #96, opened 2017-06-05
Auto Generate Update SQL?


tonypartridge:
Hi Guys,

This is looking very promising. I've been using CC for a year on and off. 

I'm struggling to see two things and this maybe my limited view after using CC. 

1. SQLi fields, so we can map to another DB to get say a list of recipes to select from another table within the component or Joomla!
2. SQL Generation, so we have to always write our own SQL. When I added the admin view with, version, meta, create,edit,save I would expect the component to generate the basic SQL needed for the fields and so on in this view?

Many thanks
Tony

Llewellynvdm:
Yes JCB does all those things and so much more, in fact that is some of the most basic things it does.

Llewellynvdm:
You can try it out, extend it your self, or purchase some of our [demo components here](http://vdm.bz/jcb-packages).

# Get Access to Video Tutorials

> **You can purchase more [demo content](http://vdm.bz/jcb-packages) for JCB including access to training & help video tutorials:**

+ *Demo Component* see the build on [github](https://github.com/namibia/demo-joomla-3-component) or get the [JCB Package](https://github.com/vdm-io/JCB-Packages/raw/master/JCB_demo.zip) _(free)_
+ *Advance Demo Component* get the [JCB Package](https://github.com/vdm-io/JCB-Packages/raw/master/JCB_demoAdvanced.zip) _([buy key](http://vdm.bz/get-advance-demo-key))_
+ *Sermon Distributor* see the build on [github](https://github.com/SermonDistributor/Joomla-3-Component) or get the [JCB Package](https://github.com/vdm-io/JCB-Packages/raw/master/JCB_sermondistributor.zip) _([buy key](http://vdm.bz/get-sermon-distributor-key))_
+ *Location Data* see the build on [github](https://github.com/vdm-io/Joomla-Location-Data) or get the [JCB Package](https://github.com/vdm-io/JCB-Packages/raw/master/JCB_locationData.zip) _([buy key](http://vdm.bz/get-location-data-key))_
+ *Help View Integration*
+ *Training & Help Videos* tutorials in the component included.

[**GET ACCESS NOW!**](http://vdm.bz/component-builder)

tonypartridge:
I've actually found most of this out now on my flight back from JAB. Just takes a bit of getting used to.

Also, can you point me in the right direction... from what I can gather so far if I make changes and then upgrade a site I need to write custom MySQL for the upgrade routine. Is that right?

[...]

PR #102, opened 2017-06-16
Auto create update SQL

Llewellynvdm:
Resolves gh-96 by adding the feature to auto create the update SQL. Also fixes gh-100 Field Length Issue.


Llewellynvdm:
@tonypartridge you are welcome to also tryout this staging branch and give us your feedback.

README.md
```diff
@@ -122,4 +122,4 @@ Component Builder is mapped as a component in itself on my local development env
 
 Come on buy me a coffee :)
  * PayPal: pay@vdm.io
- * Bitcoin: 3H35PYwikEosvcjxHXGPLS1GufZ4b4iFu3
\ No newline at end of file
+ * Bitcoin: 1Ga1WZyxC4ArSKuiFpUxdqzjfkAR1kJhZh
```
Llewellynvdm: A new Bitcoin wallet just for JCB was created.

admin/helpers/compiler/a_Get.php
```diff
@@ -2152,6 +2227,210 @@ public function setGetData($ids,$view_code)
...
+	 */
+	protected function getHistoryWatch($type, $id)
+	{
+		// quick class object to store old history object
+		$this->tmpHistory = null;
+		// Create a new query object.
+		$query = $this->db->getQuery(true);
+
+		$query->select('h.*');
+		$query->from('#__ucm_history AS h');
```
Llewellynvdm: So we depend on the history component.

ro-ot: This seems more stable then the other ideas that were on the table.

Llewellynvdm:
Yes that feature is not ready... will add it eventually. So I pushed out what I already had finished, so we can start testing it.

admin/helpers/compiler/c_Fields.php
```diff
@@ -1026,6 +1033,93 @@ private function setField($setType, $taber, &$fieldAttributes, &$name, &$typeNam
 				$fieldSet .= PHP_EOL."\t\t\t</fields>";
 				$fieldSet .= PHP_EOL."\t\t</field>";
 			}
+			// set the repeatable field
+			elseif ($typeName === 'subform')
+			{
+				// now add to the field set
```
Llewellynvdm: If Subform is correctly set as a field type in JCB it will also now work.

PR #105, opened 2017-07-20
Imporved the Adding of Custom Config Fields, added some shields.io icons to the JCB readme

Llewellynvdm:


admin/helpers/compiler/e_Interpretation.php
```diff
@@ -1243,24 +1248,30 @@ public function setupFrontendParamFields($params,$view)
...
-				$field = str_replace('required="true"', 'required="false"', $field);
-				// add to keeper array
-				$keep[] = $field;
-			}
-			else
-			{
-				// TODO add convetion to filter fields that we should not load
-				$keep[] = $field;
+				
+				$field = str_replace('display="menu"', '', $field);
```
ro-ot: This is great option, do I have it right that the **display** is like a switch that turns it on and off as to where the field will be added? **menu** and **config**

Great stuff!!

Llewellynvdm: yes :)

Issue #108, opened 2017-08-06
details_left.php rendered not to expectation. Fields missing after compile. 


Houtsmatisering:
THe fields did not have a edit order (all 0) then you only get 2 ($size) fields in the details_left.php.

		$this->tabCounter[$viewName][1] = 'Details';
			if (isset($this->layoutBuilder[$viewName]['Details'][(int) $field['alignment']][(int) $field['order_edit']]))
			{
				$size = count($this->layoutBuilder[$viewName]['Details'][(int) $field['alignment']][(int) $field['order_edit']]) + 1;
				$this->layoutBuilder[$viewName]['Details'][(int) $field['alignment']][$size] = $name;
			}
			else
			{
				$this->layoutBuilder[$viewName]['Details'][(int) $field['alignment']][(int) $field['order_edit']] = $name;
			}
			// check if publishing fields were over written
			if (in_array($name, $this->defaultFields))
			{
				// just to eliminate
				$this->movedPublishingFields[$viewName][$name] = $name;
			}

Llewellynvdm:
Give me a recommended fix :)

How do you think should we handle the build if all is 0...

We could use the line order of the fields in the modal instead, and dump the order_edit altogether... since we added it before the option to move fields around, like it is now, was ready in Joomla.

Further the reality to convert all modal repeatable fields to subform fields are very tempting indeed. I mean the subform field type is new to Joomla and works just like repeatable fields, just not in a modal, and it seems more stable to me.

Well I don't have the time right now to do this... so a quick fix for the mentioned problem should have to do for now... please share what you think will best work.

Houtsmatisering:
I like the order setting and perhaps you want them in combination with the tabs. My problem was the items did not appear and I had to debug to find out why.
I think a (dirty) fix could be if the items exists add it to the end by giving it a high edit order 999 + the layout item count. 
The user will then see the fields and can go to the edit order.

Better fix but not quick is renumber the duplicates in edit order before adding to the layout builder.

Llewellynvdm:
@Houtsmatisering please test the fix. Watch [this video](https://youtu.be/cxryVA5Cni8) for more info.

Llewellynvdm:
This is basically the fix:
```
// first fix the zero order
// to insure it lands before all the other fields
// as zero is expected to behave
if ($field['order_edit'] == 0)
{
	if (!isset($this->zeroOrderFix[$viewName]))
	{
[...]

Issue #106, opened 2017-08-04
Builder did not pas form data to server


Houtsmatisering:
Builder did not run.
Latest joomla 
Builder default php javascript change joomla.submitform( task, cform) to make sure correct form is posted. Same done for submit button.

Regards,
Jeroen

Llewellynvdm:
Hmmm try to refresh the page... ctrl F5 this will reload all JavaScript. I did not get this error so I can debug. IF refreshing the page does not help, please give more details since I am not 100% sure I understand the issue.

Houtsmatisering:
[default.php.zip](https://github.com/vdm-io/Joomla-Component-Builder/files/1202160/default.php.zip)

attached you find changes i did. when rendered the compiler html has multiple admin forms. the javascript did not specify the admin form. so it will post default. I could be wrong but it worked for me.

joomla 3.4.7 and php 7 

Jeroen

Llewellynvdm:
Wow as far as I know JCB is not ready for php7, so you got it to work in php7?

Llewellynvdm:
Okay I looked at your fix, but it causes conflict with other buttons on the page... so I wrote [a fix](https://github.com/vdm-io/Joomla-Component-Builder/commit/7a902d454acfa6d948c6295ce27be91e145ef908#diff-d04c4a3960ed08962125a268be562576L46), try out the [staging branch](https://github.com/vdm-io/Joomla-Component-Builder/tree/staging). Let me know if it works for you.

Llewellynvdm:
This seems like a [better fix](https://github.com/vdm-io/Joomla-Component-Builder/commit/fb7b5027c27585324eb172edd5cb7664056027ed#diff-d04c4a3960ed08962125a268be562576R58) please test and let me know.

Houtsmatisering:
The default.php tested ok. Building works on php 7 still checking if the created component works. Thanks.

Llewellynvdm:
@Houtsmatisering please give me feedback on php7 at [this open issue](https://github.com/vdm-io/Joomla-Component-Builder/issues/47), will safe me some time :+1:

PR #109, opened 2017-08-12
Compiler cross-browser fix, zero order fix, typo fix and improved default values.

Llewellynvdm:
- Fixed gh-106 to insure the correct form is submitted on compile, also some other minor bug fixes
- Improved the answer for gh-106 to insure the correct form is submitted on compile
- Fixed gh-108 to insure that all fields get added if multiple zeros are found. Improved the default value options in global config to allow for array of default values. Fixed typo forward slash.


ro-ot:
Can review this now, only next week if still needed.

Tony301:
I have tested this, and the compiler works in Chrome, Firefox and Safari. Also an admin view that has no ordering set on fields, works as expected. All fields are loaded.

PR #110, opened 2017-08-20
Initial JCB API

Llewellynvdm:
- Added JCB initial API frontend to run backups of JCB components
- Setup automated backup system with cronjob.
- Added email helper to send backup key to trusted email address


admin/compiler/joomla_3/Helper.php
```diff
@@ -440,12 +447,20 @@ public static function getActions($view,&$record = null,$views = null)
 	**/
 	public static function getModel($name, $path = JPATH_COMPONENT_ADMINISTRATOR, $component = '###component###')
 	{
-		// load some joomla helpers
-		JLoader::import('joomla.application.component.model');
 		// load the model file
-		JLoader::import( $name, $path . '/models' );
-		// return instance
-		return JModelLegacy::getInstance( $name, $component.'Model' );
+		JModelLegacy::addIncludePath( $path . '/models' );
```
ro-ot: This is much better!

admin/compiler/joomla_3/Helper_email.php
```diff
@@ -276,4 +276,70 @@ public static function send($recipient, $subject, $body, $textonly, $mode = 0, $
...
 		return $sendmail;
 	}
+
+	/**
+	 * Set the HTML email body
+	 *
+	 * @return  string on success
+	 *
+	 */
+	public static function setHtmlEmailBody($html, $subject)
```
ro-ot: Maybe this function name should show it is loading the html in a table row.

admin/compiler/joomla_3/Helper_site.php
```diff
@@ -331,7 +342,14 @@ public static function getVars($table, $where = null, $whereString = 'user', $wh
...
 			$query->select($db->quoteName(array($what)));
-			$query->from($db->quoteName('#_'.$main.'_'.$table));
+			if (empty($table))
+			{
+				$query->from($db->quoteName('#__'.$main));
+			}
+			else
+			{
+				$query->from($db->quoteName('#__'.$main.'_'.$table));
+			}
```
ro-ot: This will break getVars calls already in use!

admin/helpers/componentbuilderemail.php
```diff
@@ -0,0 +1,342 @@
...
+		// Create a JMail object
+		$mail = JMail::getInstance();
+
+		// check if set to global
+		if ('global' == $mailer)
+		{
+			// get the global details
+			$globalConf	= JFactory::getConfig();
+			
+			$mailer		= $globalConf->get('mailer');
```
ro-ot: There are some new values that could be used from the global mailer.

admin/compiler/joomla_3/Helper_email.php
Llewellynvdm: like `setTableHtmlBody($html, $subject)` ?

admin/compiler/joomla_3/Helper_site.php
Llewellynvdm: oops... this was an accident, I was trying something and then saw it will not work, but did not change it back. Thanks!

admin/compiler/joomla_3/Helper_email.php
ro-ot: What about?
```
setTableBody($html, $subject)
```
Short and sweet.

ro-ot:
Since this brings many new files and concepts to JCB I think the version should jump to 2.5

This will insure that all who knows how versioning works will take notice that a fairly big change is happening. I mean you are adding a front-end to JCB starting with a very basic API controller the does backups. I like what you are doing, just a lot of code doing nothing... is there not a way just to add a controller?

I mean can we not adapt JCB to be able to just build a controller if the site view and model is not even being used?

Issue #113, opened 2017-08-21
Automating Setting of Permissions


mwweb:
I posted this a while back, but it was closed without discussion, so I'm going to repost it, as I believe that this would be a major improvement to JCB.

I think a great enhancement to JCB would be the "automatic" setting of permissions upon install. This can, from my own experience, be accomplished by a two-step process. Someone might have a better implementation, but this is what worked on my 21 site views.

To extend JCB to support automatic ACL settings:

1. Add 2 new fields to JCB -> Admin Views -> Permissions repeatable field. 1 field would query and populate from #__viewlevels, the other from #__usergroups.

2. On compile of the component, the compiler would add to the compile of the permissions already being done. Right now it does similar to {} in the rules field in the assets table, until you go in and manually set every permission and save, which could be extremely frustrating for end users. The code below would set the looks and looking views on the DEMO component to automatically accessible by public on install, and create to registered.

```
               // $type (install, update or discover_install)
                if ($type == 'install') {
                        $db = JFactory::getDBO();
                        $query = $db->getQuery(true);
                        $query->update($db->quoteName('#__assets'));
                        $defaults = '{"site.looking.access":{"1":1},"site.looks.access":{"1":1},"core.create":{"2":1},"core.delete":{"2":1},"core.edit":{"2":1},"core.edit.state":{"2":1},"core.edit.own":{"2":1},"core.edit.created_by":{"2":1},"core.edit.created":{"2":1},"look.create":{"2":1},"look.delete":{"2":1},"look.edit":{"2":1},"look.edit.add":{"2":1},"look.edit.alias":{"2":1}}'; // JSON format for the parameters, see PHP serialize()
                        $query->set($db->quoteName('rules') . ' = ' . $db->quote($defaults));
                        $query->where($db->quoteName('name') . ' = ' . $db->quote('com_[[[component]]]'));
                        $db->setQuery($query);
                        $db->query();
				}
```

Any thoughts?

Note: The code above would be generated in the PHP Postflight (install) in script.php.

Llewellynvdm:
The problem we have here is that 99% the public group is ID 1, but it is a movable ID and setting it hard coded in your component could be an issue.

What we did is we added the option to remove permission control on the view, so that it does no have a permissions check. This is not ideal but it is a option, since if the view is set by default to public access, it may not even need access control.

[...]

PR #114, opened 2017-08-22
Assets Rules Improvment

Llewellynvdm:
- added site-view switch to allow default public access if access is set on the site view as explained in gh-113


admin/views/admin_views/view.html.php
```diff
@@ -149,7 +149,12 @@ protected function addToolBar()
 			{
 				JToolBarHelper::custom('admin_views.exportData', 'download', '', 'COM_COMPONENTBUILDER_EXPORT_DATA', true);
 			}
-                } 
+                }
+		if ($this->user->authorise('joomla_component.import_components'))
+		{
+			// add Import Components button.
+			JToolBarHelper::custom('joomla_components.smartImport', 'upload', '', 'COM_COMPONENTBUILDER_IMPORT_COMPONENTS', false);
```
ro-ot: This is a bug, or do you really want the import of components to show in every backend view? I must have missed this before. Sorry!

Llewellynvdm: no you are right... This is a bug!

admin/views/joomla_components/view.html.php
```diff
@@ -164,7 +159,12 @@ protected function addToolBar()
 			{
 				JToolBarHelper::custom('joomla_components.exportData', 'download', '', 'COM_COMPONENTBUILDER_EXPORT_DATA', true);
 			}
-                } 
+                }
+		if ($this->user->authorise('joomla_component.import_components'))
+		{
+			// add Import Components button.
+			JToolBarHelper::custom('joomla_components.smartImport', 'upload', '', 'COM_COMPONENTBUILDER_IMPORT_COMPONENTS', false);
```
ro-ot: Why did you move this? the buttons are now no longer next to each other.

admin/helpers/compiler/e_Interpretation.php
```diff
@@ -13664,6 +13827,13 @@ public function setAccessSections()
 					$this->langContent['admin'][$siteTitle]	= $siteName.' (Site) Access';
 					$this->langContent['admin'][$siteDesc]	= ' Allows the users in this group to access site '.ComponentbuilderHelper::safeString($siteName,'w').'.';
 					$this->componentGlobal[$sortKey]              = "\t\t".'<action name="site.'.$siteCode.'.access" title="'.$siteTitle.'" description="'.$siteDesc.'" />';
+					
+					// check if this site view requires access rule to default to public
+					if (isset($site_view['public']) && $site_view['public'] == 1)
+					{
+						// we use one as public group (TODO we see if we run into any issues)
+						$this->assetsRules[] = '"site.'.$siteCode.'.access":{"1":1}';
```
ro-ot: Nice so we can just extend on this **assetsRules** for the fields and back-ends. So it works just like the old **paramsBuilder** now **extensionsParams**. I think this will work great!

admin/views/joomla_components/view.html.php
Llewellynvdm: Well if you removed all components then you could not import any... so I moved the button to function only that does not requirer a selection, so it will show even if there is no items. I can see it messed up the lineup... :-1:  will see if we can fix that. But it has to show even if there is no items.

admin/helpers/compiler/e_Interpretation.php
Llewellynvdm: Yea it is a start!

Llewellynvdm:
@mwweb is this working for you? can I bump the version upgrade?

mwweb:
This worked fine on the demo component.

I think this is a good start on it. At least the site views can be automatically set to public. Perhaps sometime in the future we can look at, and discuss, extending this even further. An example, create and edit permissions you generally wouldn't want public, but at least registered. But that can be a start.

Issue #115, opened 2017-08-30
Add to GitHub folder


mwweb:
I'm trying to understand a bit more about HOW to use the custom github folder for multiple projects.

I have JCB option set for Git Folder Path, and JCB does create the component folder and populate it with the component, which works great for a single project.

However, if I have more than one, it doesn't work great, because if I clone my existing repo to that folder, so they are linked, JCB deletes everything in that folder before copying files.

What I found from doing some tests...

I use BitBucket for my repo, and have used them for several years. BitBucket gives 2 options: GIT or Mercurial.

If GIT is selected for the repo, it creates a .git folder. The .git folder is preserved during compile and population of the git folder.

However, if the repo is Mercurial, then it creates a .hg folder. The .hg folder gets deleted during compile and population of the git folder.

In a quick scan, it looks like helpers/componentbuilder.php function starting at 111 would need to be extended to keep .git and .hg then it would be more cross-"git" compatible.

Llewellynvdm:
You will see that each component has its own system name, this is what is used to build the git folder name. 

So that is firstly how you great various projects.

Then if you would like to have changes in the same project, lets say you want to try something but not lose all the changes made in the main branch, then simply checkout a new branch and then compile the component and it will place all the changes in the new branch and not change the main branch.

So you will need to know how to work with git.

JCB does not remove the .git folder.

Llewellynvdm:
ha sorry did not read to the end... hmmm okay so you would like to also include the .hg folder

Llewellynvdm:
Well, maybe we should add a place where you can tell JCB what not to remove during compilation. This way we can server any concept in the future. Where do you think will this best work. Globally or per project.

mwweb:
Yes it's the .hg folder, since the repo I originally setup is Mercurial, and I don't want to lose all my change logs in changing to git.

I think, perhaps, having a way of excluding files to delete might be good. I've never used it before, but I think subversion is another format that might be used, and I believe it creates a .svn folder.

I think having global would be the easiest, but on a per project basis it would be more flexible. The could be differences between the different projects.

Llewellynvdm:
Still not sure I understand 

> However, if I have more than one, it doesn't work great, because if I clone my existing repo to that folder, so they are linked, JCB deletes everything in that folder before copying files.

Do you mean clone from the remote repo? if so why would you clone from the remote repo, unless you are more then one developer on the project.

[...]

PR #119, opened 2017-09-13
Upgrade to FTP feature. Dynamic Excluding on repository.

Llewellynvdm:
- Moved FTP signature to its own table, so we can use it more securely around JCB (even in the global settings). Improved the Linked view option in admin-views.
- Resolve gh-115 to allow excluding of folders and files in the repository folder


mwweb:
I will test this later this evening.

mwweb:
I had to create a new test repo, since over this past weekend I went though a process of converting all my Mercurial to Git.

In my new test repo, which is Mercurial, I add .hg to the field in the component settings, and on compile it left the .hg folder intact, and added the component folders and files to the directory.

It would appear that this is working as intended.

Llewellynvdm:
Great thanks for testing this!

Issue #120, opened 2017-09-14
error in export component


sepehr09:
hi,
i have JCB 2.5.4 (last version until today) and when i'm trying to export component ( joomla components-> Export component) the following error is displayed: 

An error has occurred.
1054 Unknown column 'fa' in 'where clause'

also i made multilingual component (language translations for "fa-IR" language) Could this be the reason for the error?

What is the solution?
thank you very much

Llewellynvdm:
Hi I have been thinking about this issue for the last two days... I am not sure really.

I don't know why this is hapening... hmmm I need to have it before me... lets see Unknown column.

Let my try debug this a few more days and see if I can find something. I have never had this issue before.

Just out of interest what version of PHP do you have installed and on what operating system?

Llewellynvdm:
Okay I found the bug :) will push a fix!

sepehr09:
thank you very much for you answer & follow up

PR #121, opened 2017-09-18
Fixed gh-120 that allows export of component with language translations

Llewellynvdm:


admin/helpers/compiler/a_Get.php
```diff
@@ -1773,19 +1773,8 @@ public function getFieldData($id,$name_single = null,$name_list = null)
 				$field->xml = $this->setDynamicValues(json_decode($field->xml));
 
 				// load the type values form type params
-				$properties = json_decode($field->type_properties,true);
-				unset($field->type_properties);
+				$field->properties = json_decode($field->properties,true);
```
ro-ot: Now that was way to easy!

Llewellynvdm: Yea :+1:  I was relieved... I do not want to change more field for now... it just to big a risk. So I did these and lets see if there is any unforeseen results. So we can keep track of the all round reality of changing from a repeatable field to a subform field type. It really effects everything in relation to automation.

Llewellynvdm:
Thank you!

Issue #136, opened 2017-10-10
Input variables exceeded 1000 (to big table sets in JCB)


Llewellynvdm:
I am moving all repeatable fields to subforms in JCB as most of you know by now. But I have run into an issue that may cause some headaches to say the least.

Repeatable fields were stored in a single field as json, but the subforms are not, this means when a page has many subforms the variables skyrocket... and this is the case with Admin Views, not to even mention Joomla Component Views. So now we run into the following error soon as we start to really use these subforms.

```
PHP Warning:  Unknown: Input variables exceeded 1000.
To increase the limit change max_input_vars in php.ini.
```
### Option One
This seems like a case to decouple many of the tabs (consepts) from the Admin and Joomla Component Views. So that we will have smaller tables and smaller forms. I mean the Joomla Component Table has nearly 200 columns....

### Option Two
The other option is that all those who use JCB increase the `max_input_vars` to about 5000 at least. Yet this is usually discourage as bad practice.

### The implication of option one
- Will really bump my time to do this... this is not a good time for this
- Will make JCB even more complex then before
- Will do amazing scalability and faster page load
- Will remain easy to run JCB on shared hosting
### The implication of option two
- Will be a quick fix
- Will break good practice of micro components
- Will not run well on shared environments
- I really hate this option.... but wow, I am even more upset at Joomla for removing something that worked, **Repeatable Fields**.

> I am almost finished with the next update that would have removed repeatable fields from site_view, custom_admin_view and admin_view. Then during testing I opened some admin views and when trying to save it... landed back on the dashboard. Looking at the logs I discoverer the above error... so now the big question.

# Does any of you know how we can better resolve this issue?

kevincam3:
I wish I was that good of a PHP programmer to have something to offer. I run JCB only on my localhost so for now I have no problem with just bumping up the max_input_vars

mwweb:
I, too, only run JCB on my localhost. For ease, since there are other things that Joomla has depreciated and will be removed in J!4 (see my notes on issue 126 about JRequest and JError being depreciated). It actually looks like a lot of the legacy functions may be getting removed in J!4, as part of a cleanup of the code.

For now, at least, it might be best to _REQUIRE_ the increase of max_input_vars to 5000, for the initial major upgrade, perhaps with a notice upon install or update about the requirement, but that this "issue" will be resolved in a future release.

[...]

PR #138, opened 2017-10-12
Removing repetable fields from site view, custom admin view, and admin view. 

Llewellynvdm:
- Also updated the compiler, removed permissions on fields to speedup page load in admin and joomla component views.
- Working on resolving gh-136 issue.


mwweb:
I tested. Simply installing into a new site, everything appears to be fine. However, I installed on a clone of my dev site, attempted to open a large admin view that I have, and got an out of memory errors, and I have my PHP memory set to 256MB. 

Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 63889187 bytes) in /var/www/dev/plugins/system/debug/debug.php on line 1225

I increased memory_limit to 512M and it loaded, but took forever to do so. I'm guessing that the high memory usage is due to the number of input vars that it's having to load/process.

Llewellynvdm:
I am wondering if we should not port the repeatable field into the JCB core... I mean if Joomla stops support for it... we can still use it without them. That field was a powerhouse...

> I will start by decoupling the fields and conditions. 

__Lets do a test @mwweb if you delete line 82 in component->views->admin_view->tmpl->edit.php is there a dramatic improvement in the page load?__

Llewellynvdm:
I have do so and the page speed is back to normal... so I think we move those two concepts out onto its own view/table. Please confirm your side.

ro-ot:
After seeing the tremendous slow page load, I agree we must move the fields out. But still display the fields with some details via ajax call in the fields tab, with buttons to edit.

@Llewellynvdm let me know what I can help with.

Hey we had to drop the repeatable field, it was just a buggy nightmare in the Javascript. The direction you are going will work, and make JCB even better believe me.

mwweb:
I removed that line, and the view does load quickly once again.

On my big view, it loaded, but only with the Settings tab. all the others were missing. But, it looks like this may be due to a jQuery error, likely due to the size of the view (number of fields, tabs, etc.)

Uncaught RangeError: Maximum call stack size exceeded

I think I'm off to sleep. It's almost 3am here.

Llewellynvdm:
@ro-ot I need help with the ajax... I will setup the new fields view. Not sure what to call it yet. But okay... I will start on that right away. Soon as I am ready I will push the changes.

ro-ot:
Maybe just call it `admin_fields` and `admin_field` with no direct menu link, only via the admin views. Okay once you have all in place I will dump the ajax in. My M-JCB is out of sync with yours... so maybe I just give you the code and you add it.

admin/helpers/compiler/a_Get.php
```diff
@@ -1208,191 +1209,259 @@ public function getAdminViewData($id)
...
 				{
-					$field['settings'] = $this->getFieldData($field['field'],$name_single,$name_list);
+					$field['settings'] = $this->getFieldData($field['field'], $name_single, $name_list);
+				}
+			}
+			unset($view->addfields);
+			// set the conditions
+			$view->addconditions = json_decode($view->addconditions,true);
+			// check what type of conditions array we have here (should be subform... but just incase)
+			// This could happen due to huge data sets
```
ro-ot: Stunning!

Llewellynvdm:
Okay @mwweb & @ro-ot  I have decoupled the fields and conditions. Please test and give feedback.

## Note this is not production ready!

> I still need to change the JCP package export/import area... as this is now totally broken. But the compiler should work, and so should the rest of JCB.

Llewellynvdm:
@ro-ot You will see there is a placeholder for the fields to load via ajax... if you can just build the PHP function that returns the ready build HTML to the ajax call then I will do the rest.

**So the table from which to get the fields is admin_fields and the table to get the conditions is admin_fields_conditions**

mwweb:
OK. "Smaller" view load fine, and it looks like the new fields and conditions load fine.

However, on a large view with lots of fields and conditions, I'm still getting a jQuery error:

jquery error

At some point, we may want to look at reorganizing the layout of the settings tab:

admin layout

Llewellynvdm:
Some times you need to just do a ctrl-f5 page refresh to get the newest javaScript and stuff. I don't see those errors.

Any suggestions on the page layouts?

mwweb:
It actually looks like it may be related to the number of fields. From the DB:

admin_fields

mwweb:
In doing some digging:

In MySQL, the text field type allows a maximum 65535 characters. If it's over that, then the error above. The suggestions were to use largetext, or even better medium or large blob.

I'm not sure if that will resolve the issue or not though. I manually changed addfields to largetext, and the error isn't there, but still getting the jQuery error.

Llewellynvdm:
Are you saying a text column (in the DB) is to small? Wow how many field do you have?

O are you taking about a view in your component? Well if that is the case... yes you need to decoupled :+1:

mwweb:
I'll do more tests later. I emptied the fields in the DB, and "error" still persists, even after a ctrl f5.

Llewellynvdm:
In what view are you seeing this? in the admin view? or in the admin fields view?

Llewellynvdm:
I still have to clean out some scripts from the admin_view

mwweb:
One question, and should probably be somewhere else...

In this build, the new hidden views: admin_fields and admin_fields_conditions are showing as a single view in admin/components/com_componentbuilder/views. By default, JCB requires Name (Single Record) and Name (List of Records). In this version you only have a Single Record (no list of records). Was that accomplished via custom code, or some other way?

Llewellynvdm:
It is a default JCB behavior if a number of things lineup. Then I also made the field readonly, since if all the things lineup the correct view is always selected. So the user does not need to select it.

### What must lineup? 
Well the sort answer is, the field name is "admin_view" and the referral view is "admin_view" and it always pass and referral ID in the url. So the first time you open the admin_fields view it uses the `ref=admin_view` and `refid=1` to automatically select the correct view from the list. JCB will always do that, but only on the first load when it has no values set, and has not been saved. Since I know each admin_view should only have one admin_fields linked to it... I allow this automation and then with the readonly switch in the field prevent the user from changing it. Hey it can be bypassed, but then the user will know that it broke because of him/her bypassing this automation.

Llewellynvdm:
There is so many things like that... that i have forgotten of some of them already and need to open old component to see how I did it... then I realize **hey** JCB already has that trick in place... lol so I end up learning JCB...

Llewellynvdm:
This last push has the new layout :+1:

mwweb:
To answer the question: In what view are you seeing this? in the admin view? or in the admin fields view?, it is in admin view that i am seeing the jquery error, and the message about the field length was seen in admin_view, admiin_view_fields, and admin_views_conditions

Llewellynvdm:
I think we should work on a limiting factor around how many fields should be allowed per/view at this time it is set to 800, but that is way to many... I think around 200 is a more reasonable number.

I mean we must realize that 200 fields can mean 200 columns in the table and that is bad practice.

I have pushed Joomla to its limits with JCB and it worked with repeatable fields, but this subform is not allowing this huge amounts of values to work well. 

We have now completely isolated the fields to its own table/view and still it can't handle so many fields. 

I don't know what more can be done, without resorting to custom scripting (outside of Joomla) You see I have not added any script to the page besides the Joomla default scripts... and if they give errors we are probably going to cut back on the number of fields allowed per view.

Llewellynvdm:
Test this last push... I am off to bed it is 6am here. Was a long hard coding night for me. The storm is almost over... @ro-ot please bump that script for me.

mwweb:
The new layout looks good.

However, I think there is something more that is causing that jQuery Range Error. I just tested with one of my smaller views (25 fields, including the default fields the JCB adds such as id, created, published, etc), and I got the range error. Out of 14 views, most with around 25 or less fields, only 4 opened without any errors.

I can't see that I have done anything "abnormal", except for adding fields and tabs, which is normal.

ro-ot:
@mwweb I also do not seem to get these errors. Are you :100:% sure that your browser cache has been cleared. Also looking at your error message, it seems you JavaScript is being cached, so also clear your server cache, or just turn it off.

ro-ot:
@Llewellynvdm I think you did a great Job! 

### Regarding the allowed fields size. I have these thoughts.

- Do not change the field limitation
- Instead add a warning once a user goes beyond two hundred fields
- This warning should simply state that adding more then a 200 fields is bad practice.
- Lets change the DB field size to _MEDIUMTEXT_ for both fields and conditions

### Now for loading the fields back via Ajax

- I will for now only return text (Table) of the fields and the settings
- We may want to convert this into a Ajax form (to allow changes without going to the admin fields page)
- So we will start very basic, just so users can see the fields

> I like the new layout and the edit/create notice when you click the new fields button.

mwweb:
I will dig into it more today, but the error occurred on  different sites i set up on my localhost. And occurred on a copy of the site i uploaded to my web server. But like i said, I'll dig into it more today.

mwweb:
I can't say definitively, but it may have to do with something in the data conversion from repeatable fields to subforms. I just went and recreated two of my views that were throwing that jquery range error. I recreated them with identical tabs, permissions, fields, conditions, everything.

The result? No errors.

ro-ot:
Can you give us the two json objects? The one done by the JCB conversion and the one done when you manually saved the subform?

mwweb:
Definitely differences in the 2 JSON's.

`Original

{"addfields0":{"field":"1597","list":"1","order_list":"1","title":"1","alias":"0","sort":"1","search":"1","filter":"0","link":"1","tab":"1","alignment":"1","order_edit":"1","permission":"0"},"addfields1":{"field":"23","list":"1","order_list":"2","title":"0","alias":"1","sort":"0","search":"0","filter":"0","link":"0","tab":"1","alignment":"1","order_edit":"2","permission":"0"},"addfields2":{"field":"1574","list":"0","order_list":"0","title":"0","alias":"0","sort":"0","search":"0","filter":"0","link":"0","tab":"1","alignment":"1","order_edit":"3","permission":"0"},"addfields3":{"field":"913","list":"0","order_list":"0","title":"0","alias":"0","sort":"0","search":"0","filter":"0","link":"0","tab":"1","alignment":"1","order_edit":"4","permission":"0"},"addfields4":{"field":"1503","list":"0","order_list":"0","title":"0","alias":"0","sort":"0","search":"0","filter":"0","link":"0","tab":"2","alignment":"1","order_edit":"1","permission":"0"},"addfields5":{"field":"1501","list":"0","order_list":"0","title":"0","alias":"0","sort":"0","search":"0","filter":"0","link":"0","tab":"2","alignment":"1","order_edit":"2","permission":"0"},"addfields6":{"field":"1502","list":"0","order_list":"0","title":"0","alias":"0","sort":"0","search":"0","filter":"0","link":"0","tab":"2","alignment":"1","order_edit":"3","permission":"0"}}

New

{"addfields0":{"field":"1597","list":"1","order_list":"1","title":"1","sort":"1","search":"1","link":"1","tab":"1","alignment":"1","order_edit":"1","permission":"0"},
"addfields1":{"field":"23","list":"1","order_list":"2","alias":"1","tab":"1","alignment":"1","order_edit":"2","permission":"0"},"addfields2":{"field":"1574","order_list":"0","tab":"1","alignment":"1","order_edit":"3","permission":"0"},"addfields3":{"field":"913","order_list":"0","search":"1","tab":"1","alignment":"1","order_edit":"4","permission":"0"},"addfields4":{"field":"1503","order_list":"0","tab":"2","alignment":"1","order_edit":"1","permission":"0"},"addfields5":{"field":"1501","order_list":"0","tab":"2","alignment":"1","order_edit":"2","permission":"0"},"addfields6":{"field":"1502","order_list":"0","tab":"2","alignment":"1","order_edit":"3","permission":"0"}}`

mwweb:
OK. I think it's in conjunction with the differences in the addfields and addconditions when converted, and also to the below.

I tested this by copying the content from the "recreated" fields and conditions to the old, and still had the error. I then saw that the only major difference was with addtables in admin_view. The old had contents, the new didn't. With the different JSON in place, and deleting the contents of addtables, it started working.

In the admin_view table, there is a field called "addtables". This field is populated in the one throwing an error, but is not populated in the one that isn't. As a test, I went into the one that is throwing the error and emptied "addtables".

No errors any more.

ro-ot:
The only difference I see is that some values are missing in the second. This could be because @Llewellynvdm changed the field type from radio to checkbox for some of these fields, and it only has one value 1 not zero. This could be the issue.

mwweb:
It's more than just the new tables admin_view_fields and admin_view_conditions, as in my testing that alone didn't resolve the error. Once I did a compare of the 2 records in admin_view, I saw that addtables was populated in the original record, but not the new. I deleted from the original record, and it started working.

So, something between the new tables and the addtables looks like it may be the culprit. Also, not sure that it's needed any longer, but in admin_view the addfields and addconditions fields are still there.

I have a JCB Package of the component, from JCB 2.5.5, that I can provide if that would make it easier to debug.

Llewellynvdm:
From my own testing I also found that not all conversions were done, so I added a few fixes around JCB to deal with this. 

I think we can move forward with this as it clearly is simply huge datasets that are to big to convert all at once. So adding little conversion scripts in the model everywhere where these fields gets loaded is what I have done. So that even after the install if a convention was not done completely it will be done the moment you open the view.

mwweb:
I'll test new version when available.

Llewellynvdm:
About the fields still being in the database, yes we will only finally remove them few update further, just until we are sure all is working as planned. So I am not clearing them out just yet.

ro-ot:
I hope this will work. You will need to add the controller values. But from the update you send me, this is the php for the Ajax area

```
protected $functionArray = array(
			'field' => 'setFieldsNames',
			'list' => 'setYesNo',
			'title' => 'setYesNo',
			'alias' => 'setYesNo',
			'sort' => 'setYesNo',
			'search' => 'setYesNo',
			'filter' => 'setYesNo',
			'link' => 'setYesNo',
			'permission' => 'setYesNo',
			'tab' => 'setTabName',
			'alignment' => 'setAlignmentName',
			'target_field' => 'setFieldsNames',
			'target_behavior' => 'setTargetBehavior',
			'target_relation' => 'setTargetRelation',
			'match_field' => 'setFieldsNames',
			'match_behavior' => 'setMatchBehavior');

protected function getSubformTable($idName, $oject)
{
	if (ComponentbuilderHelper::checkJson($oject) && ComponentbuilderHelper::checkString($idName))
	{
		$array = json_decode($oject, true);
		if (ComponentbuilderHelper::checkArray($array))
		{ 
			// Build heading
			$head = array();
			foreach ($array as $headers)
			{
				foreach ($headers as $header => $value)
				{
					$head[$header] = '<th>' . ComponentbuilderHelper::safeString($header, 'Ww');
				}
			}
			// build the rows
			$rows = array();
			if (ComponentbuilderHelper::checkArray($array) && ComponentbuilderHelper::checkArray($head))
			{
				foreach ($array as $nr => $values)
				{
					foreach ($head as $key => $t)
					{
						// set the value for the row
						if (isset($values[$key]))
						{
							$this->setSubformRows($nr, $this->setSubformValue($key, $values[$key]), $rows);
						}
						else
						{
							$this->setSubformRows($nr, $this->setSubformValue($key, ''), $rows);
						}
					}
				}
			}
			// build table
			if (ComponentbuilderHelper::checkArray($rows) && ComponentbuilderHelper::checkArray($head))
			{
				// return the table
				return $this->setSubformTable($head, $rows);
			}
		}
	}
	return false;
}

protected function setSubformTable($head, $rows)
{
	$table[] = "<div class=\"row-fluid\">";
	$table[] = "\t<div class=\"subform-repeatable-wrapper subform-table-layout subform-table-sublayout-section-byfieldsets\">";
	$table[] = "\t\t<div class=\"subform-repeatable\">";
	$table[] = "\t\t\t<table class=\"adminlist table table-striped table-bordered\">";
	$table[] = "\t\t\t\t<thead>";
	$table[] = "\t\t\t\t\t<tr>";
	$table[] = "\t\t\t\t\t\t". implode("", $head);
	$table[] = "\t\t\t\t\t</tr>";
	$table[] = "\t\t\t\t</thead>";
	$table[] = "\t\t\t\t<tbody>";
	foreach ($rows as $row)
	{
		$table[] = "\t\t\t\t\t<tr class=\"subform-repeatable-group\">";
		$table[] = "\t\t\t\t\t\t" . $row;
		$table[] = "\t\t\t\t\t</tr>";
	}
	$table[] = "\t\t\t\t</tbody>";
	$table[] = "\t\t\t</table>";
	$table[] = "\t\t</div>";
	$table[] = "\t</div>";
	$table[] = "</div>";
	// return the table
	return implode("\n", $table);
}

protected function setSubformValue($header, $value)
{
	if (array_key_exists($header, $this->functionArray) && method_exists($this, $this->functionArray[$header]))
	{
		$value = $this->{$this->functionArray[$header]}($header, $value);
	}
	// if no value are set
	if (!ComponentbuilderHelper::checkString($value))
	{
		$value = '-';
	}
	return $value;
}

protected function setSubformRows($nr, $value, &$rows)
{
	// build rows
	if (!isset($rows[$nr]))
	{
		$rows[$nr] = '<td>'.$value.'</td>';
	}
	else
	{
		$rows[$nr] .= '<td>'.$value.'</td>';
	}
}

protected $fieldsArray = array(
			'admin_fields' => 'addfields',
			'admin_fields_conditions' => 'addconditions');

public function getFieldsDisplay($type)
{
	if (isset($this->fieldsArray[$type]))
	{
		// get the view name & id
		$values = $this->getViewID();
		// check if we are in the correct view.
		if (!is_null($values['a_id']) && $values['a_id'] > 0 && strlen($values['a_view']) && $values['a_view'] === 'admin_view')
		{
			// get the field data
			if ($fieldsData = ComponentbuilderHelper::getVar($type, (int) $values['a_id'], 'admin_view', $this->fieldsArray[$type]))
			{
				// decode the fields
				return $this->getSubformTable($type, $fieldsData);
			}
		}
		return '<div class="control-group"><div class="alert alert-info">' . JText::sprintf('No %s have been linked to this view. Soon as this is done it will be displayed here.', ComponentbuilderHelper::safeString($type, 'w')) . '</div></div>';
	}
	return '<div class="control-group"><div class="alert alert-error"><h4>' . JText::_('Type Error') . '</h4><p>' . JText::_('There has been an error, if this continues please inform your system administrator of a type error in the fields display request!') . '</p></div></div>';
}

protected function setAlignmentName($header, $value)
{
	switch ($value)
	{
		case 1:
			return JText::_('Left in Tab');
		break;
		case 2:
			return JText::_('Right in Tab');
		break;
		case 3:
			return JText::_('Full Width in Tab');
		break;
		case 4:
			return JText::_('Above Tabs');
		break;
		case 5:
			return JText::_('Underneath Tabs');
		break;
		case 6:
			return JText::_('Left of Tabs');
		break;
		case 7:
			return JText::_('Right of Tabs');
		break;
	}
	return JText::_('not set');
}

protected $fieldNames = array();

protected function setFieldsNames($header, $value)
{
	$bucket = array();
	if (ComponentbuilderHelper::checkArray($value))
	{
		foreach ($value as $field)
		{
			if (!isset($this->fieldNames[$field]))
			{
				if (!$this->fieldNames[$field] =  ComponentbuilderHelper::getVar('field', (int) $field, 'id', 'name'))
				{
					$this->fieldNames[$field] = JText::_('No Field Found');
				}
			}
			$bucket[] = $this->fieldNames[$field];
		}
	}
	elseif (is_numeric($value))
	{
		if (!isset($this->fieldNames[$value]))
		{
			if (!$this->fieldNames[$value] =  ComponentbuilderHelper::getVar('field', (int) $value, 'id', 'name'))
			{
				$this->fieldNames[$value] = JText::_('No Field Found');
			}
		}
		$bucket[] = $this->fieldNames[$value];		
	}
	// return found fields
	if (ComponentbuilderHelper::checkArray($bucket))
	{
		return implode('<br />', $bucket);
	}
	return JText::_('No Field Found');
}

protected $tabNames = array();

protected function setTabName($header, $value)
{
	if (ComponentbuilderHelper::checkArray($this->tabNames))
	{
		// get the view name & id
		$values = $this->getViewID();
		if (!is_null($values['a_id']) && $values['a_id'] > 0 && strlen($values['a_view']) && $values['a_view'] === 'admin_view')
		{
			if ($tabs = ComponentbuilderHelper::getVar('admin_view', $values['a_id'], 'id', 'addtabs'))
			{
				$tabs = json_decode($tabs, true);
				if (ComponentbuilderHelper::checkArray($tabs))
				{
					$nr = 1;
					foreach ($tabs as $tab)
					{
						if (ComponentbuilderHelper::checkArray($tab) && isset($tab['name']))
						{
							$this->tabNames[$nr] = $tab['name'];
							$nr++;
						}
					}
				}
			}
		}
	}
	// has it been set
	if (ComponentbuilderHelper::checkArray($this->tabNames) && isset($this->tabNames[$value]))
	{
		return $this->tabNames[$value];
	}
	return JText::_('Details');
}

protected function setYesNo($header, $value)
{
	if (1 == $value)
	{
		return JText::_('Yes');
	}
	return JText::_('No');
}

protected function setTargetBehavior($header, $value)
{
	switch ($value)
	{
		case 1:
			return JText::_('Show');
		break;
		case 2:
			return JText::_('Hide');
		break;
	}
	return  JText::_('not set');
}

protected function setTargetRelation($header, $value)
{
	switch ($value)
	{
		case 1:
			return JText::_('Isolate');
		break;
		case 2:
			return JText::_('Chain');
		break;
	}
	return  JText::_('not set');
}

protected function setMatchBehavior($header, $value)
{
	switch ($value)
	{
		case 1:
			return JText::_('Is (only 4 list/radio/checkboxes)');
		break;
		case 2:
			return JText::_('Is Not (only 4 list/radio/checkboxes)');
		break;
		case 3:
			return JText::_('Any Selection (only 4 list/radio/checkboxes/dynamic_list)');
		break;
		case 4:
			return JText::_('Active (only 4 text_field)');
		break;
		case 5:
			return JText::_('Unactive (only 4 text_field)');
		break;
		case 6:
			return JText::_('Key Word All case-sensitive (only 4 text_field)');
		break;
		case 7:
			return JText::_('Key Word Any case-sensitive (only 4 text_field)');
		break;
		case 8:
			return JText::_('Key Word All case-insensitive (only 4 text_field)');
		break;
		case 9:
			return JText::_('Key Word Any case-insensitive (only 4 text_field)');
		break;
		case 10:
			return JText::_('Min Length (only 4 text_field)');
		break;
		case 11:
			return JText::_('Max Length (only 4 text_field)');
		break;
		case 12:
			return JText::_('Exact Length (only 4 text_field)');
		break;
	}
	return  JText::_('not set');
}
```
The view id function is already in the Ajax class. Please feel free to make any changes you feel is needed.

Llewellynvdm:
Wow stunning, thanks! I was wondering if I should ask again.

ro-ot:
I have tested it at it works great.

The import and export still broken.

But the admin view works very well now.

admin/helpers/compiler/e_Interpretation.php
```diff
@@ -2875,6 +2868,13 @@ public function setCustomViewDisplayMethod(&$view)
 				$method .= PHP_EOL.PHP_EOL."\t\t//".$this->setLine(__LINE__)." set the document";
 				$method .= PHP_EOL."\t\t\$this->setDocument();";
 			}
+
+			$method .= PHP_EOL.PHP_EOL."\t\t//".$this->setLine(__LINE__)." Check for errors.";
+			$method .= PHP_EOL."\t\tif (count(\$errors = \$this->get('Errors')))";
+			$method .= PHP_EOL."\t\t{";
+			$method .= PHP_EOL."\t\t\tthrow new Exception(implode(".'"\n", $errors), 500);';
```
ro-ot: So the new error handler is in place. I see you also updated all the files :+1:

admin/models/ajax.php
```diff
@@ -436,19 +458,34 @@ public function getFieldsDisplay($type)
...
 				// get the field data
 				if ($fieldsData = ComponentbuilderHelper::getVar($type, (int) $values['a_id'], 'admin_view', $this->fieldsArray[$type]))
 				{
-					// decode the fields
-					return $this->getSubformTable($type, $fieldsData);
+					// get the table
+					$table = $this->getSubformTable($type, $fieldsData);
+					// set notice of bad practice
+					$notice = '';
+					if (isset($this->rowNumber) && $this->rowNumber > 50)
```
ro-ot: Only 50? is that not to low?

Llewellynvdm: Well it is only a notice and in reality they can add up to 800. I just found the moment you go beyond that number the new field views really slowdown. So telling them that they should try to stay below 50 is a good idea. 

Do you think we need to explain more in terms of why it is bad practice?

Llewellynvdm:
@mwweb can you give me any feedback on those errors if they are now resolved with this last few updates. It will be helpful, thanks!

mwweb:
My apologies, I have been traveling. I just did a reinstall back to 2.5.5, then installed the most recent build. From what I can see, by opening each view, everything is losing correctly.

Llewellynvdm:
I should have the final release out today! I am almost done with the import of JCB packages to ensure backward compatibility. So when I release it, I would need testing to check backward compatibility. I mean to insure the packages that were exported in the past will import and end in the right place.

Llewellynvdm:
I am still having issue on the import of New JCB packages. @ro-ot can you take a look at this and see what I missed?

ro-ot:
@Llewellynvdm the JCB package export seems to be the issue. Not all the items are being exported, it seems like some items are missed.

That is all I have for now.

Llewellynvdm:
Okay now we have the new export and import working. Still need to do some checks on old packages. Would appreciate some feed back on this last push.

Issue #147, opened 2017-10-25
Improvement: Sort order if fields in Editing the Admin View 


kevincam3:
It would be nice to sort the fields in the order I set them to display in the admin view in the attached view:

![image](https://user-images.githubusercontent.com/870934/31999475-1eb3cf94-b972-11e7-9baf-d2e876cd001d.png)

Llewellynvdm:
The automated sort of the fields has been removed since you can with the subform drag and drop the fields in the correct order very easy.

To avoid us change you own selected order, we removed the feature.

kevincam3:
So do we need the "Order in Edit" if Drag and Drop works?

Llewellynvdm:
yes the drag and drop does work :+1: 

You still need to set the order for edit and order for list, as those orders are for the compiler. The drag and drop is just for you... :)

kevincam3:
Ok, so what I was thinking is we just have one and not both sorts. Unless there's a reason that someone would want them to be different.

kevincam3:
It's not a big deal it was just something that was counter intuitive to me that after I changed the sort order in "Order edit" I was expecting to see them listed in that order after I saved it.

Llewellynvdm:
I do not have one view where those are the same. So of course we want them both, they target two different concepts. The one for the list layout and the other for the edit layout in a tab and position.

Llewellynvdm:
Hey we can always add the sort feature back.. if you are up to the task.

[Look here is the old code](https://github.com/vdm-io/Joomla-Component-Builder/blob/v2.5.5/admin/models/admin_view.php#L2822) that sorted the fields on save. 

Convert it to now work on the new admin_fields. This is [the issue that we last addressed this](https://github.com/vdm-io/Joomla-Component-Builder/issues/43).

[This is the area (line 848 in the admin_fields model) the code must go](https://github.com/vdm-io/Joomla-Component-Builder/blob/master/admin/models/admin_fields.php#L848).

So adapt the code, test it, and once you think it works as you expected, share it in this issue. Then I will test it and add it to the core once we are sure it works.

kevincam3:
I thought it was just a simple matter of a query that was returning the results in this image and we would just sort by "Order edit" column. But if it's more complicated than that we can leave it as is.

![image](https://user-images.githubusercontent.com/870934/31999475-1eb3cf94-b972-11e7-9baf-d2e876cd001d.png)

Llewellynvdm:
[...]

Issue #146, opened 2017-10-24
Uncaught RangeError: Maximum call stack size exceeded


mwweb:
I'm still doing testing with 2.5.8, since I wanted to ensure that everything is working fine.

In JCB 2.5.5 everything would compile fine, and everything would load properly for my component on the frontend and the backend.

After the upgrade to JCB 2.5.8 I was doing some testing and was getting Error 0, which is such a non-descriptive error message. In looking at the call stack, it looks like under JCB 2.5.8 it was having issues with the joined com_categories table:

```

1 | () | JROOT/libraries/joomla/database/driver/mysqli.php:660
-- | -- | --
2 | JDatabaseDriverMysqli->execute() | JROOT/components/com_mostwantedrealestate/models/properties.php:324
3 | MostwantedrealestateModelProperties->getCatidIdCeef_B() | JROOT/components/com_mostwantedrealestate/models/properties.php:249

```

In looking at the dynamic get, the joined field is there, but I tested deleting then re-adding the table, which is when I noticed that it wasn't populating the selection area. Some other views were loading.

Digging deeper, there is a jQuery issue present on the admin views. I mentioned this during the staging, but it looks like it is still an issue. This issue is present on localhost, and if I backup and install the dev on my web hosting server.
```
Uncaught RangeError: Maximum call stack size exceeded
    at Function.Sizzle (jquery.js?31617fba9b2d24f9efd1ac0ebd420dd2:844)
    at Function.jQuery.find (jquery-migrate.js?31617fba9b2d24f9efd1ac0ebd420dd2:290)
    at jQuery.fn.init.find (jquery.js?31617fba9b2d24f9efd1ac0ebd420dd2:2826)
    at jQuery.fn.init.jQuery.fn.find (jquery-migrate.js?31617fba9b2d24f9efd1ac0ebd420dd2:677)
    at jQuery.fn.init (jquery.js?31617fba9b2d24f9efd1ac0ebd420dd2:2949)
    at new jQuery.fn.init (jquery-migrate.js?31617fba9b2d24f9efd1ac0ebd420dd2:241)
    at jQuery (jquery.js?31617fba9b2d24f9efd1ac0ebd420dd2:75)
    at HTMLDocument.<anonymous> (index.php?option=com_componentbuilder&view=dynamic_get&layout=edit&id=92:196)
    at fire (jquery.js?31617fba9b2d24f9efd1ac0ebd420dd2:3232)
[...]

Issue #145, opened 2017-10-24
JCB Layouts in dynamicGet


mwweb:
I know you have been swamped with all the changes, but I wanted to make a suggestion on moving about some of the subforms on views. For example, on the dynamic gets you can see how it looks:
![jcb-layouts](https://user-images.githubusercontent.com/19194012/31919436-556b6d82-b817-11e7-9bad-11164c22e11f.png)

Everything is scrunched together, making it difficult to see what entries there are. Perhaps just moving them out of the right in tab to left in tab, under "selection". That would give full width for the subform. Or a settings tab like you did on admin views.

ro-ot:
What screen size are you working on? that is not how it looks on my screen.

ro-ot:
![image](https://user-images.githubusercontent.com/2594383/31919754-d5809e64-b864-11e7-8b3f-0296076b326d.png)
This is on my screen, so it actually looks good.

mwweb:
The recommended resolution of my graphics card of 3000x2000.

Llewellynvdm:
Sure we can look at the layout... improvements are always good.

I just what to take small steps at a time, to avoid losing those who already are using the JCB component.

Issue #144, opened 2017-10-22
created date show in wrong tab after update


sepehr09:
hi & thanks for new update,
before this version we select tab 15 to showing created date field in default tab but after update tab 15 is not more existing & how we can show created date in default tab? ( i wan to show created date in admin listview)
![image](https://user-images.githubusercontent.com/11753906/31860014-0017487a-b721-11e7-82d9-cc5c035f8846.png)

best regards!

Llewellynvdm:
You, sure you can still setup tabs. Please watch this tutorial https://youtu.be/NFp_CtE0LZI

Llewellynvdm:
ahh, I now see what you really asking.... great stuff, with the next update I will add that option back in. Sorry :) I missed that, but hey now there is one more tutorial :+1:

Llewellynvdm:
I was looking for a quick work around and it seems we are stuck.... I am on a difficult place I have already fixed this issue, but since I also started moving the **Joomla Component View** away from the repeatable fields like yesterday..... my local _JCB as a mapped component_ has diverged from the working package so drastically that I can not do a quick update. I have only two options, one is to wait until the next release that other is to give you some custom code to place in the compiler to do a temporal work around... how urgent are you for this? can you wait a week?

sepehr09:
Thanks for your attention, i'm glad we have you!
if you can send me custom code It will be very good but if you are busy right now i can wait.
If any help I can do, I'll be happy to do it!
thanks.

Llewellynvdm:
Okay I found a work around, that does not need any code.

Just add a tab, named `publishing` all lowercase. then add that field `created` to that tab and it should work. 

Let me know :+1:

Llewellynvdm:
Just not that this workaround will only work when you replace the default fields so you can use them in the list view, or what to tweak the field in some way. With the coming release you will be able to add other fields to the publish tab as well as the default once. So it will work as before only better.

sepehr09:
yes that's work! actually i'm test adding tab before but the key is all words must be lowercase!

PR #152, opened 2017-10-26
Major Upgrade away from repeatable fields in Joomla component view.

Llewellynvdm:
- Added status, copy and delete sync between admin_view linked views (admin_fields, admin_fields_conditions).
- Added notice of language (tanslations) added or not added with details to compiler
- Resolved gh-146 compiler error on joined db in dinamic get thanks to @mwweb & @ro-ot
- Resolved gh-147 by adding the sort of fields back into the save method
- Resolved gh-144 to ensure that the published tab (fields overwriting and adding) option is available again.
- Resolved gh-145 by moving the subforms to their own tab in dynamic get view
- Converted all repeatable fields to subform fields in Joomla component view
- Moved 9 subforms and other fields to their own table and view (decopuling them fom Joomla component view), that means we added 9 more views and tables to JCB
- Added all the ajax for buttons and display views to Joomla component view
- Added tmp scripts all across the new areas with subforms to ensure all repeatable fields are converted. Will be removed in v2.7.0
- Added synced copy, change state and delete in Joomla components view in relation to all tables linked to it (same as with admin views)


ro-ot:
I tried to copy a Joomla Component view

```
Catchable fatal error: Object of class stdClass could not be converted to string in /home/root/libraries/src/Filter/InputFilter.php on line 345
```
So it seems there is a error in your copy method.

mwweb:
My preliminary testing this build appears to be working correctly.

Llewellynvdm:
No the build/compiler does not work. **CAUTION** the compiler has not been updated.

Llewellynvdm:
Just the data have been moved and the views setup and all the nice **UI** stuff :+1:

mwweb:
I realize that. I'll test the compiler when it's ready. I tested by adding a publishing tab, and placed a field in the publishing tab. I will see if it properly places that field in the publishing tab when compile is working, which would resolve one of the issue suggestions.

Llewellynvdm:
Ya all the last few issues where addressed in the compiler, but not the transition to subforms in the joomla_component view.

I think this release is going to be the biggest yet with over 35k new lines of code.

mwweb:
With such major changes, is just a minor incremental version change appropriate? With such major changes, would it, possibly, be more appropriate to have this version as 2.6.0? Then work up from there, then for the Joomla! 4 version, start at 3.0? Just thinking out loud on this. I went around and around with the Joomla Devs on something similar when they released J! 1.6 (which they admitted later should have been 2.0, which is what I was saying to them).

Llewellynvdm:
I think the release should be v2.6.0

Then the compiler and all will be ready and working. So yes we should give a clear heads-up on this release. I pushed it to v2.5.9 but I was intending to bump to 2.6 before we let this one out :+1:

Llewellynvdm:
Okay I have managed to do the complete update, everything should now work as expected.

- The whole new Joomla Component View is ready
- The compiler is ready
- The export of JCB packages is ready
- The import of JCB package

@ro-ot  & @mwweb please test and let me know if you find any issues.

mwweb:
I ran a test and it compiled fine. The custom field that I added to tab "publishing" is showing on the Publishing tab in the admin view. One thing I think we may want to point out of people is that if they want to add to that tab, they need to keep the position to Left in Tab. 

Export and Import of the component appears to be working as well.

Llewellynvdm:
Try adding it to the right and then compile the component, you will see a message :)

mwweb:
I did that initially and there was a message that the position didn't exist. That's where i moved it and recompiled.

Llewellynvdm:
Yes, is that not enough?

Llewellynvdm:
Can you do some test on exporting and importing of JCB packages?

mwweb:
I guess we can leave the note as is. If there is any confusion/questions we can add to the note later. I dobt know how many will use the ability to add to the publishing tab. 

I tested exporting and importing one component, and it appears to have worked flawlessly. I'll create a whole new install today and test more.

mwweb:
OK. I did a complete export of my components, and imported into a brand new test site. It looks like FTPs didn't export, and out of 37 layouts only 1 is in the new site.

Llewellynvdm:
Okay the FTP don't export by default, as a security measure.

I will check the layouts issue that is weird.

Thanks!

Llewellynvdm:
It is weird because it is in an area I did not change in this update... I just tested it again my self with the Sermon Distributor component and it moved all the layouts as expected. I will keep looking...

Just that you know the export will only export layouts linked in **templates** or **site_views** or **custom_admin_views** or other layouts of course that are linked to components. IF not linked to components it will not export it. Not that I think this is the case with you... just want to make sure we are clear on that implementation principle.

Llewellynvdm:
@mwweb have you been able to resolve the issues you mentioned. We are about ready to let this upgrade merge with the core.

mwweb:
I just ran another test. The component that I am exporting is my component that is now "live". There are multiple admin views, several site views, several templates being called in the site views, and several layouts being called from the templates.

Once again, it looks like all the admin views, sites views, and templates are being exported, but not the layouts.

I event tested with doing an export of Sermon Distributor, then importing. The same thing happens: only Series Table Row imports.

Llewellynvdm:
This is strange, I am not seeing the same issue at all, in fact this JCB seems to work more effectively then any of the version before.

Would you mind giving me the Sermon Distributor JCB package that you have exported. So I can try import it on my blank install to see what happens.

mwweb:
Here you go.

[JCB_sermondistributor.zip](https://github.com/vdm-io/Joomla-Component-Builder/files/1428424/JCB_sermondistributor.zip)

Llewellynvdm:
Can you please email me the key?

mwweb:
I had to dig through all my old emails to find your email address. I just sent the package and key to your joomla email address.

Llewellynvdm:
I still did not receive your email... but I found the issue. Please test and let me know.

Just **note** the problem was with the exporter not the importer :+1:

Llewellynvdm:
I was not seeing the error before, since I was using old exporter JCB packages. But I think it is resolved now.

Llewellynvdm:
We are about ready to publish this.... just need you to verify that the issue is resolved.

mwweb:
That looks like it resolved the issue.

mwweb:
I need to run 1 more test for you. When attempting to compile, I received an error: 0 Call to undefined function bcadd()

mwweb:
Yes, it looks like there is an error in the compiler in regards to bcadd.

jcb_error

mwweb:
Update on the above error: the error ONLY occurs in PHP 7, not in PHP 5. So, there appears to be something that is not compatible with PHP 7 in the compiler.

Llewellynvdm:
This is a function for calculations that should be turned on in PHP.
http://php.net/manual/en/function.bcadd.php

PR #159, opened 2017-11-01
Minor Fixes (lets try again)

Llewellynvdm:
Here we can push a few fixes for version 2.6.0


mwweb:
It looks like it is working fine.

PR #186, opened 2017-11-24
Prevent getItem from clearing $item->params

stutteringp0et:
Additional check added to params test to ensure that if parent::getItem values are not removed if present.

Pull Request for Issue # .

### Summary of Changes

Added !is_array($item->params) test to getItem method.

### Testing Instructions

I'm bringing additional fields into my form via ajax and saving that data as param values.  Before the patch, param data is lost.  After the patch, param data remains.

### Expected result

param data remains

### Actual result

param data is lost

This should be a transparent change for most users, as there isn't a facility in JCB to create param values in the forms.  It takes something weird like what I'm doing to notice it's an issue.


Llewellynvdm:
Thanks!

PR #191, opened 2017-12-03
String Manipulation to SimpleXML and DOMDocument

stutteringp0et:
I removed the string manipulation in favor of something more developer friendly.  The drawback is that the resulting XML is not as pretty, although I did run the processed fields through DOMDocument pretty printer.  The entire resultant XML could be run through a pretty printer, but still won't achieve the tab indented node attributes.

The entire process could be made much more efficient, but this was just an initial conversion.

### Summary of Changes

Removed string manipulation in favor of direct xml manipulation

### Testing Instructions

generate a form

### Expected result

a form is generated

### Actual result

a form is generated

### DIFF
--- a/admin/helpers/compiler/c_Fields.php
+++ b/admin/helpers/compiler/c_Fields.php
@@ -1,28 +1,28 @@
 <?php
 
-/**--------------------------------------------------------------------------------------------------------|  www.vdm.io  |------/
-    __      __       _     _____                 _                                  _     __  __      _   _               _
-    \ \    / /      | |   |  __ \               | |                                | |   |  \/  |    | | | |             | |
-     \ \  / /_ _ ___| |_  | |  | | _____   _____| | ___  _ __  _ __ ___   ___ _ __ | |_  | \  / | ___| |_| |__   ___   __| |
-      \ \/ / _` / __| __| | |  | |/ _ \ \ / / _ \ |/ _ \| '_ \| '_ ` _ \ / _ \ '_ \| __| | |\/| |/ _ \ __| '_ \ / _ \ / _` |
-       \  / (_| \__ \ |_  | |__| |  __/\ V /  __/ | (_) | |_) | | | | | |  __/ | | | |_  | |  | |  __/ |_| | | | (_) | (_| |
-        \/ \__,_|___/\__| |_____/ \___| \_/ \___|_|\___/| .__/|_| |_| |_|\___|_| |_|\__| |_|  |_|\___|\__|_| |_|\___/ \__,_|
-                                                        | |                                                                 
-                                                        |_| 				
-/-------------------------------------------------------------------------------------------------------------------------------/
+/* * --------------------------------------------------------------------------------------------------------|  www.vdm.io  |------/
+  __      __       _     _____                 _                                  _     __  __      _   _               _
+  \ \    / /      | |   |  __ \               | |                                | |   |  \/  |    | | | |             | |
+  \ \  / /_ _ ___| |_  | |  | | _____   _____| | ___  _ __  _ __ ___   ___ _ __ | |_  | \  / | ___| |_| |__   ___   __| |
+  \ \/ / _` / __| __| | |  | |/ _ \ \ / / _ \ |/ _ \| '_ \| '_ ` _ \ / _ \ '_ \| __| | |\/| |/ _ \ __| '_ \ / _ \ / _` |
+  \  / (_| \__ \ |_  | |__| |  __/\ V /  __/ | (_) | |_) | | | | | |  __/ | | | |_  | |  | |  __/ |_| | | | (_) | (_| |
+  \/ \__,_|___/\__| |_____/ \___| \_/ \___|_|\___/| .__/|_| |_| |_|\___|_| |_|\__| |_|  |_|\___|\__|_| |_|\___/ \__,_|
+  | |
+  |_|
+  /-------------------------------------------------------------------------------------------------------------------------------/
 
-	@version		2.6.0
-	@created		30th April, 2015
-	@package		Component Builder
-	@subpackage		compiler.php
-	@author			Llewellyn van der Merwe <http://www.vdm.io>
-	@my wife		Roline van der Merwe <http://www.vdm.io/>	
-	@copyright		Copyright (C) 2015. All Rights Reserved
-	@license		GNU/GPL Version 2 or later - http://www.gnu.org/licenses/gpl-2.0.html 
-	
-	Builds Complex Joomla Components 
-                                                             
-/-----------------------------------------------------------------------------------------------------------------------------*/
+  @version		2.6.0
+  @created		30th April, 2015
+  @package		Component Builder
+  @subpackage		compiler.php
+  @author			Llewellyn van der Merwe <http://www.vdm.io>
+  @my wife		Roline van der Merwe <http://www.vdm.io/>
+  @copyright		Copyright (C) 2015. All Rights Reserved
+  @license		GNU/GPL Version 2 or later - http://www.gnu.org/licenses/gpl-2.0.html
+
+  Builds Complex Joomla Components
+
+  /----------------------------------------------------------------------------------------------------------------------------- */
 
 // No direct access to this file
 defined('_JEXEC') or die('Restricted access');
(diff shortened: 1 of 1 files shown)


admin/helpers/compiler/c_Fields.php
```diff
@@ -1,2238 +1,2315 @@
 <?php
 
-/**--------------------------------------------------------------------------------------------------------|  www.vdm.io  |------/
```
Llewellynvdm: ca not allow this spacing it will mess everything up, please check your IDE to use the same spacing as the repo.

admin/helpers/compiler/c_Fields.php
```diff
@@ -1,2238 +1,2315 @@
...
+    /**
+     * set the Field set of a view
+     * 
+     * @param   array    $view  The view data
+     * @param   string   $component The component name
+     *
+     * @return  string The fields set in xml
+     * 
+     */
+    public function setFieldSet($view, $component) {
```
Llewellynvdm: all over your spacing is messing up the pull request!!!

Llewellynvdm:
I tried review your changes but since your spacing is a total mess I can't do any review for almost the whole file changed. First fix the spacing....

Llewellynvdm:
So that only the area you worked on shows as changed in git.

Llewellynvdm:
This is ridiculous 
image

Llewellynvdm:
If you can not with the change produce the same neat as you call it *pretty print* result, I can tell you now this is not getting merged. Sorry. I need the xml to look hand written, and therefore perfectly neat ad in order with comments and clean.

I spend tone of hours to achieve that, so if you can improve, then do. I for one can not see why string manipulation is any less developer friendly? if what it produce is more superior and clean.

stutteringp0et:
I put a lot of work into it, so I'm not giving up.  I can tell you put a lot of work into the string manipulation, but the idea of having to run multiple tests to determine if you should close a tag seems like a waste of time when the PHP xml classes do that automatically.

So.  I think I should first address the issue you have with my code formatting.  I tab my way to my desired indention level and when I'm done I use the Source > Format utility of NetBeans.  I can adjust the settings, but I need to know what the standard is.  It looks like you use an 8 character tab indent.  Is that correct?  I'll adjust my IDE settings to put function opening curly braces on the 2nd line....  Do you have your code style prescribed somewhere?

As far as reaching the "hand written" output look you want - I think I can do an XSLT transform on the finalized XML to achieve that.  I've never used XSLT for that, but what the heck.

stutteringp0et:
About the XML comments - I did take special care to preserve them in their original configurations and locations (indention fixes withstanding)

Llewellynvdm:
My best advice to you is copy the area you worked on out of the file, then hard reset and fix the indentation of what you are doing to yes **8 per tab** and only add the changes back without formatting the whole file.

I really do not want to go over work that is just spacing issues. I will want to just look at your work, not indentation stuff.

I am open to do things better, as long as it does not take us back in aesthetics nor in functionality required for behaviors and checks.

I feel like this change is going to cause my more work... but okay let me see.

Llewellynvdm:
I am trying to keep to the Joomla Code style as described in the [contributer guide](https://github.com/vdm-io/Joomla-Component-Builder/blob/master/.github/CONTRIBUTING.md).

Llewellynvdm:
Yet I know there are areas in the compiler that has not followed that, so it needs refactoring. But for now that must wait... there is to much other things for me to do.

Llewellynvdm:
you did not do what I said so the file stil has way to many changes sorry...

stutteringp0et:
I'm holding off on any more changes to this PR until I hear from you.

I think the indention of the inserted XML would be best fixed by running through tidy before writing the xml output file.  If you accept this PR, I'll modify the file write operation to tidy before write.

Llewellynvdm:
Please add commenting to the new functions... let do that in the files where most functions have their commenting in place.

I also think you will have to fix the indentation before.

admin/helpers/compiler/c_Fields.php
```diff
@@ -1622,18 +1672,19 @@ private function setFieldAttributes(&$field, &$viewType, &$name, &$typeName, &$m
...
 	 * @param   string   $viewName  The singel view name
 	 *
 	 * @return  void
-	 * 
+	 *
 	 */
 	protected function setUniqueNameKeeper(&$field, &$typeName, &$name, $viewName)
 	{
 		// setup a default field
+		$xml = new SimpleXMLElement($field['settings']->xml);
```
ro-ot: This is not a good idea, the string from JCB is not always valid xml!!!!

admin/helpers/compiler/c_Fields.php
```diff
@@ -1308,6 +1357,7 @@ private function setFieldAttributes(&$field, &$viewType, &$name, &$typeName, &$m
 		// setup a default field
 		if (ComponentbuilderHelper::checkArray($field['settings']->properties))
 		{
+			$xml = new SimpleXMLElement($field['settings']->xml);
```
ro-ot: This is not a good idea, the string from JCB is not always valid xml!!!!

stutteringp0et:
OK, I fixed indent and commented the functions

admin/helpers/compiler/c_Fields.php
stutteringp0et: If you're going to require users to insert code - shouldn't you require them to insert valid code?

ro-ot: I think in this areas we should stay with the old string manipulation to insure more freedom here for the xml typos.

ro-ot:
Hey commenting is always expected, only @Llewellynvdm gets away, and not for long :+1:

stutteringp0et:
I wasn't going to spend time adding comments until there was a light at the end of the tunnel and I was certain it wasn't a train.  The initial reception was a bit cold.  I'm glad the formatting fixes warmed things up a bit.  This will open the door for a lot of future improvements.

admin/helpers/compiler/c_Fields.php
ro-ot: That was the reason in the first place if I remember correctly. We said we rebuild the xml and do not use it directly to avoid invalid xml getting used. Indirectly allowing the user to be more sloppy and then things will still work. 

So changing this will cause some users to get odd errors.

ro-ot:
Hey noting cold here just been us two for long time, hope you understand the love :100:

ro-ot:
Always room for more, but you have to admit you had a made a mess of the formatting lol. Did you get my comments surrounding those two areas?

stutteringp0et:
I didn't see that.  What's the formatting standard?  I think I had my IDE set to the Joomla standard...but I did re-install Linux last month (upgraded storage to a mirrored SSD hardware RAID array - and it's stupid fast!)

admin/helpers/compiler/c_Fields.php
Llewellynvdm: hmmm I hate these kind of things. I mean sure both of you have a point and both are right. I know as an engineer we makes thinks stronger then what is needed to insure it works even when things go wrong. That is resilience to avoid breaking, and it is what makes JCB so powerful. So sure lets in those areas remain resilient.

Since it is not just about typos but also about the php being added in the custom field. I have not tested that but I can imagine it may strip things. @stutteringp0et have you tested these changes on custom fields with PHP and ##'s in between the "..."

ro-ot:
Just ran a test on a huge project of mine, did blank install, installed these version and then as expected:
```
String could not be parsed as XML
```
So we have a problem with those PHP strings in the XML.

stutteringp0et:
Could you post the offending xml?

ro-ot:
That is my point there is thousands of fields in my component and which one is it now? Do you have the time to search a thousand fields, I do not.

Llewellynvdm:
Please check this [info](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests) to give us the maintainers permission to make changes to this code. Since I have tried to help you here in making some changes but don't seem to succeed in pushing it up. Keeps on creating a new branch... maybe I am doing something wrong. Do you know how I can contribute to his pull request?

ro-ot:
found one:
```
<field 
	type="text"
	name="worldzone" // selfde as die databasis
	label="Worldzone"
	size="40"
	maxlength="50" 
	default=""
	description="The zone this country belongs to"
	class="text_area"
	filter="INT"
	required="false"
	message="Error! Please add zone code here."
	hint="only number"
/>
// hey moet asblief nie die een verander nie!!!
```
I left some notes in here for other team members, this is a great example of why I think @Llewellynvdm is right to request you take just two small steps back.

You did a great job on the other area why not work with us on this. I am telling you the call here to remain resilience is the right one.

stutteringp0et:
I'll change it back to string search, but I have to question the purpose of storing it as xml if you can't parse it as xml.

Why xml, why not ini, or csv key value pairs or just random crap with key value pairs injected randomly?

As far as 1000s of fields, yes - I would find the malformed xml immediately.  It might take 10 minutes to write the code to find it.

ro-ot:
Really? can we just avoid taking this personal, sure we could add checks to detect which fields are not valid xml.

I think Llewellyn chose the xml layout since it looks familiar to how fields are done in Joomla, I read the code and therefore know those values are being grabbed one at a time. So I started taking advantage of the freedoms. Don't let this become a big deal to you.

This is really not so big a deal, it is just in all my test I ran with **your code** my fields changed there names (so I had install issues with database mismatch) that I don't think this is the time and place to take on the world. 

Not wile we are in the middle of a huge upgrade, don't know if you are aware of the coming changes.

I will keep this improvement in the back of my mind, and who knows when the time is right we can take this up again.

Please just not now.

stutteringp0et:
I'm not trying to make it personal - I often use sarcasm

You can still have notes - just make them compliant with xml:

    <field 
 	type="text"
 	name="worldzone" name:note="selfde as die databasis"
 	label="Worldzone"
 	size="40"
 	maxlength="50" 
 	default=""
 	description="The zone this country belongs to"
 	class="text_area"
 	filter="INT"
 	required="false"
 	message="Error! Please add zone code here."
 	hint="only number"
     />
    

If the system stores XML - I don't want to parse text, I want to query the XML

Updated the PR to use string search to allow for malformed XML

stutteringp0et:
@Llewellynvdm - Requested changes have been made

stutteringp0et:
screenshot from 2017-12-03 17-02-20

stutteringp0et:
noted

Issue #111, opened 2017-08-20
onContentPrepare fix


Llewellynvdm:
I would like to add onContentPrepare to each text area and text field loaded in a site view, there is already some code that does this
```
// Make sure the content prepare plugins fire on description.
$data->description = JHtml::_('content.prepare',$data->description);
```
[on line 1514](https://github.com/vdm-io/Joomla-Component-Builder/blob/master/admin/helpers/compiler/e_Interpretation.php#L1514) in the Interpretation.php

But it seems to stop working, so now I would like to redo it like the way the article component implements this.

```
// Process the content plugins.
JPluginHelper::importPlugin('content');
$dispatcher->trigger('onContentPrepare', array ('com_content.article', &$item, &$item->params, $offset));
```
[on line 187](https://github.com/joomla/joomla-cms/blob/staging/components/com_content/views/article/view.html.php#L187) in the view.html.php file of the article view.

Any ideas will be helpful!

stutteringp0et:
I do a lot of work with plugins.

The first option in the config array becomes the $context variable in plugin.  This allows the plugin author to selectively apply their changes to only items that match a context.  Normally, you'd set the valid component and view dot path and let the cards fall where they may - that is, if a plugin has a narrow $context - it might not show the plugin alterations.  For context though, you can go any direction you want - broad, narrow - I saw one extension that used com_component.view.layout in its plugin context...

So, in order to do this the right way, the component name and view name must be available when this code is generated (as opposed to faking another context, which I've also seen developers do).  Passing the item by reference allows the  plugin to make changes - otherwise the plugin can look, but not touch.  I'd pass the item params as well, although the plugin would have to know what it was looking for to use them.

I've issued a PR, it should be linked by now

stutteringp0et:
Right, there is something that isn't working right - apparently ###view### isn't available at this point in e_Interpretation.php - so the closest context I can get is com_###component###

stutteringp0et:
I'm adding some additional code to create a better $context based on the available fileContentStatic keys.  So if you want to modify the code to include the view name, it will happen in this code automatically.

stutteringp0et:
Nevermind about the context, I have it.  Removing the additional logic to construct a context - I can get it from the model.

PR #192, opened 2017-12-04
Implement onContentPrepare

stutteringp0et:
Pull Request for Issue #111  .

### Summary of Changes

Implement onContentPrepare

### Testing Instructions

The easiest test is to create an editor field and within it type an email address.  If the Email Cloaking plugin is enabled, the output should have a cloaked email address.  

### Expected result

it works

### Actual result

it works

tell people it works?

### DIFF
--- a/admin/helpers/compiler/e_Interpretation.php
+++ b/admin/helpers/compiler/e_Interpretation.php
@@ -1654,12 +1654,6 @@ class Interpretation extends Fields
 	{
 		$fieldUikit = '';
 		$runplugins = false;
-		$context = array();
-		foreach(array('###component###','###view###') as $key) {
-			if(array_key_exists($key,$this->fileContentStatic)) {
-				$context[] = $this->fileContentStatic[$key];
-			}
-		}
 		foreach ($checker as $field => $array)
 		{
 			if (strpos($get['selection']['select'], $field) !== false)
@@ -1670,7 +1664,7 @@ class Interpretation extends Fields
 					$runplugins .= PHP_EOL."\t".$tab."\t".'$dispatcher = JEventDispatcher::getInstance();';
 				}
 				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field;
-				$fieldUikit .= PHP_EOL."\t".$tab."\t".'$dispatcher->trigger("onContentPrepare",array("com_'.implode('.',$context).'",&'.$string.'->'.$field.',$item->params));';
+				$fieldUikit .= PHP_EOL."\t".$tab."\t".'$dispatcher->trigger("onContentPrepare",array($this->_context,&'.$string.'->'.$field.',$item->params));';
 				// only load for uikit version 2 (TODO) we may need to add another check here
 				if (2 == $this->uikit || 1 == $this->uikit)
 				{
--- a/admin/helpers/compiler/e_Interpretation.php
+++ b/admin/helpers/compiler/e_Interpretation.php
@@ -1660,7 +1660,7 @@ class Interpretation extends Fields
 			{
 				// build decoder string
 				if(!$runplugins) {
-					$runplugins  = PHP_EOL."\t".$tab."\tJPluginHelper::importPlugin('content');\n";
+					$runplugins  = PHP_EOL."\t".$tab."\tJPluginHelper::importPlugin('content');";
 					$runplugins .= PHP_EOL."\t".$tab."\t".'$dispatcher = JEventDispatcher::getInstance();';
 				}
 				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field;
--- a/admin/helpers/compiler/e_Interpretation.php
+++ b/admin/helpers/compiler/e_Interpretation.php
@@ -1658,8 +1658,12 @@ class Interpretation extends Fields
 			if (strpos($get['selection']['select'], $field) !== false)
 			{
 				// build decoder string
-				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field." (TODO)";
-				$fieldUikit .= PHP_EOL."\t".$tab."\t".$string."->".$field." = JHtml::_('content.prepare',".$string."->".$field.");";
+				if(!$runplugins) {
+					$runplugins  = PHP_EOL."\t".$tab."\tJPluginHelper::importPlugin('content');\n";
+					$runplugins .= PHP_EOL."\t".$tab."\t".'$dispatcher = JEventDispatcher::getInstance();';
+				}
+				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field;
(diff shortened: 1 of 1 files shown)


admin/helpers/compiler/e_Interpretation.php
```diff
@@ -1653,13 +1653,18 @@ public function setCustomViewFieldDecode(&$get,$checker,$string,$code,$tab = '')
...
 			{
 				// build decoder string
-				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field." (TODO)";
-				$fieldUikit .= PHP_EOL."\t".$tab."\t".$string."->".$field." = JHtml::_('content.prepare',".$string."->".$field.");";
+				if(!$runplugins) {
+					$runplugins  = PHP_EOL."\t".$tab."\tJPluginHelper::importPlugin('content');\n";
+					$runplugins .= PHP_EOL."\t".$tab."\t".'$dispatcher = JEventDispatcher::getInstance();';
+				}
+				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field;
+				$fieldUikit .= PHP_EOL."\t".$tab."\t".'$dispatcher->trigger("onContentPrepare",array($this->_context,&'.$string.'->'.$field.',$item->params));';
```
Llewellynvdm: Why did you add `$item->params` to every field check?

admin/helpers/compiler/e_Interpretation.php
```diff
@@ -1653,13 +1653,18 @@ public function setCustomViewFieldDecode(&$get,$checker,$string,$code,$tab = '')
...
+		$runplugins = false;
 		foreach ($checker as $field => $array)
 		{
 			if (strpos($get['selection']['select'], $field) !== false)
 			{
 				// build decoder string
-				$fieldUikit .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field." (TODO)";
-				$fieldUikit .= PHP_EOL."\t".$tab."\t".$string."->".$field." = JHtml::_('content.prepare',".$string."->".$field.");";
+				if(!$runplugins) {
+					$runplugins  = PHP_EOL."\t".$tab."\tJPluginHelper::importPlugin('content');\n";
```
Llewellynvdm: Why have a new line `\n` please remove, and make sure to only use `PHP_EOL` to insure it works on windows.

Llewellynvdm: You realize it will place it in a loop... I mean this code is place inside the loop of items. So it will load the dispatcher ever time around. I think we must move this dispatcher out of the loop.

admin/helpers/compiler/e_Interpretation.php
stutteringp0et: Each trigger is the plugin running on a different field.  The plugin event doesn't retain params from the previous plugin run - so it's included every time the plugin event is called.

admin/helpers/compiler/e_Interpretation.php
stutteringp0et: I inserted multiple editors into my test and the dispatcher was created only once.  Even if it's created more than once, it won't make a difference to the dispatcher - it's a singleton, it's the same object.

newline removed

Llewellynvdm:
I have merged it into staging branch, next time please do target the master branch but the staging, thanks!

admin/helpers/compiler/e_Interpretation.php
Llewellynvdm: Yes I understand that, but you don't understand that the params variables are not aways on the page. So to pass them without a check is bound to give errors. 

This onContnentPrepare is used in the dynamicGets and as you know the user is in full control of what data is retrieved. So if there is not params best just pass a null.

I will fix this, just that you know in all we do in compiler we are never sure about any value, we must always check before we use value.

Llewellynvdm: Even the `_context` value is not always there since we are moving between different modules `JModelList` and `JModelAdmin` and `JModelItem` and every one deals with `_context` in its own way. Like in `JModelList` context is found in [$this->context](https://github.com/Llewellynvdm/joomla-cms/blob/staging/libraries/src/MVC/Model/ListModel.php#L37) not `$this->_context` so we must check in what model this code is going before we decide the context implementation.

stutteringp0et: Your tables always create a params field (I haven't found a way to turn it off).  but now that I look at the dynamic get, it appears that param isn't one of the fields imported by default.  It should be simple enough to (property_exists($item,'params')?$item->params:$this->params) - returning the component params when the item params don't exist.  The plugin is happy to get a params object and you have no error.

$this->context vs $this->_context - That sounds like an inconsistency to be resolved.  Without a context, the plugin activity will be unpredictable.  Some plugins don't care, but if someone wanted to write a plugin to act upon ONLY the output of this JCB extension - they wouldn't be able to identify JCB content.  The context is kind of important.

Here's an example of a plugin operating only on a specific context - this one alters the alias to a base65 conversion on the weblink id - so another plugin can use that alias to create a shortcode:

    public function onContentAfterSave($context, $article, $isNew)
    {
        if ( ($context == "com_weblinks.weblink") && ($isNew === true) ) {

            $db = JFactory::getDbo();
            $query = $db->getQuery(true);

            $anyBase = new \weblinks\AnyBase();
            $alias = $anyBase->encode($article->id);

            $fields = array(
                'alias = "' . $alias . '"'
            );

            $conditions = array(
                'id = ' . $article->id
            );

            $query->update($db->quoteName('#__weblinks'))->set($fields)->where($conditions);

            $db->setQuery($query);
            $db->execute();
        }
    }

Llewellynvdm: I have resolved this issues, will push it up to stage soon. I just mentioned these things to help you look deeper into the complexity of all being done.

Here is a little peak at the fix that is coming
```
public function setCustomViewFieldonContentPrepareChecker(&$get, $checker, $string, $code, $tab = '', $context, $params)
{
	$fieldPrepare = '';
	$runplugins = false;
	foreach ($checker as $field => $array)
	{
		if (strpos($get['selection']['select'], $field) !== false)
		{
			// build decoder string
			if(!$runplugins) {
				$runplugins = PHP_EOL.$tab."\t//".$this->setLine(__LINE__)." Load the JEvent Dispatcher";
				$runplugins .= PHP_EOL.$tab."\tJPluginHelper::importPlugin('content');";
				$runplugins .= PHP_EOL.$tab."\t".'$this->_dispatcher = JEventDispatcher::getInstance();';
			}
			$fieldPrepare .= PHP_EOL."\t".$tab."\t//".$this->setLine(__LINE__)." Make sure the content prepare plugins fire on ".$field;
			$fieldPrepare .= PHP_EOL."\t".$tab."\t\$".$field." = new stdClass();";
			$fieldPrepare .= PHP_EOL."\t".$tab."\t\$".$field.'->text =& '.$string.'->'.$field.'; // value must be in text';
			$fieldPrepare .= PHP_EOL."\t".$tab."\t".'$this->_dispatcher->trigger("onContentPrepare",array('.$context.',&$'.$field.','.$params.'));';
		}
	}
	// load dispatcher
	$this->JEventDispatcher = array('###DISPATCHER###' => ($runplugins?:''));
	// return UIKIT fix
	return $fieldPrepare;
}
```

Issue #188, opened 2017-11-28
default fields added to all views are named incorrectly in the display


kevincam3:
I just noticed that in the screenshot there's "date_created" but in the xml it's "created", I also didn't find "date_modified" but "modified" I guess there could be others.

![image](https://user-images.githubusercontent.com/870934/33331490-e1998fda-d448-11e7-904a-be51fcb9d341.png)

kevincam3:
Related to this, it didn't override the value. I got a warning that the value already exists:

![image](https://user-images.githubusercontent.com/870934/33333046-00f4b82e-d44d-11e7-9774-68b827fded78.png)

Llewellynvdm:
It looks like I did not update the note, it should say created and modified and not date_created... I willl update the note.

So if you want to override those fields to just change the name to created or modified.

kevincam3:
@Llewellynvdm,

Any idea why the override isn't working properly when I try to install?

Llewellynvdm:
Kevin this can be many reasons, best investigate your database and then the update sql in the Joomla package. I can not say, just that it is not normal and most likely an error on your part.

The normal implementation of this feature still works as expected, so to debug this will be up to you. You are welcome to give us feedback if you find out more, and if you feel this is a bug. In which case you will need to give much more information and explanation from your side.

PR #195, opened 2017-12-10
Library manager implementation

Llewellynvdm:
Pull Request for many Issues

### Summary of Changes
- Added interface of the new decoupled libraries concept gh-92 
- String Manipulation to SimpleXML and DOMDocument (thanks to @stutteringp0et)
-  Improved the get snippets area to load one library at a time. started on the adaptation of the compiler that is needed for the new libraries
- Resolve gh-111Implement onContentPrepare (thanks to @stutteringp0et)
- Renamed the uikit lib folders.
- Resolved the gh-188 by updating the note.
- Resolved gh-92 by implementation of the library manager in the compiler.
- Resolve gh-181 to insure all urls added to xml is converting special characters with htmlspecialchars.

### Testing Instructions
- Test the new library implementation in all related views
- Check the fields xml files created (to test the new SimpleXML and DOMDocument )
- When adding a library to a site view and see if the library details is added to the **view.html.php**


Llewellynvdm:
@stutteringp0et will also be great if you can run some test on this pull request.

admin/compiler/joomla_3/JViewLegacy_list.php
```diff
@@ -197,9 +197,12 @@ protected function addToolBar()
 	 */
 	protected function setDocument()
 	{
-		$document = JFactory::getDocument();
-		$document->setTitle(JText::_('COM_###COMPONENT###_###VIEWS###'));
-		$document->addStyleSheet(JURI::root() . "administrator/components/com_###component###/assets/css/###views###.css");
+		if (!isset($this->document))
+		{
+			$this->document = JFactory::getDocument();
```
ro-ot: bold move, but okay makes it more.

admin/helpers/compiler/f_Infusion.php
```diff
@@ -318,6 +318,7 @@ protected function buildFileContent()
 						&& ComponentbuilderHelper::checkString($this->customScriptBuilder['php_document'][$viewName_single]))
                                         {
                                                 // ###DOCUMENT_CUSTOM_PHP### <<<DYNAMIC>>>
+						$this->customScriptBuilder['php_document'][$viewName_single] = str_replace('$document->', '$this->document->', $this->customScriptBuilder['php_document'][$viewName_single]);
```
ro-ot: ok this will help avoid breaking things, thanks!

admin/helpers/compiler/e_Interpretation.php
```diff
@@ -2560,6 +2609,11 @@ public function setCustomViewCustomItemMethods(&$main_get,$code)
 						{
 							$decoder_filter = $this->setCustomViewFieldDecodeFilter($get,$decodeFilter,'$item','$items[$nr]',$default['code'],"\t\t");
 						}
+						$contnetprepare = '';
```
ro-ot: spelling typo: **$contentprepare**

admin/helpers/compiler/e_Interpretation.php
```diff
@@ -12402,9 +12697,14 @@ public function setConfigFieldsets($timer = 0)
 				$dbkey = 'g';
 				foreach ($this->componentData->config as $field)
 				{
-					$xmlField = $this->setDynamicField($field, $view, $viewType, $lang, $viewName, $listViewName, $spacerCounter, $placeholders, $dbkey, false);
-
-					if (ComponentbuilderHelper::checkString($xmlField))
+					$newxmlField = $this->setDynamicField($field, $view, $viewType, $lang, $viewName, $listViewName, $spacerCounter, $placeholders, $dbkey, false);
+					// tmp hack untill this whole area is also done in xml (TODO)
```
ro-ot: @stutteringp0et hope you see this :+1: we need to extend the new SimpleXML change into the config area.

Llewellynvdm: Well that is not going to be so simple. If it was I could have done it with this upgrade. Reality is that many of the compiler functions is based on the string manipulation way and changing it to xml will break many things specially in relation to the config area.

I am grateful for the work you did, but I know that the full impact of the change has not been realized. So there is not much more to be done, the way the compiler works already makes it easy to make a change in one place and have it impact the whole compiler simply because of tones of polymorphism and encapsulation.

So here is the deal, to adapt the xml aspect into the config area is very simple in relation to changing the fields to SimpleXML objects, but the hard work is to identify all the areas where this adaptation will impact the logic of the original string manipulation implementation.

None of you realized this, not @ro-ot nor @stutteringp0et and so naturally when I notice some of my components not adding its config fields and that the whole config and menu area was dead, I added this little hack, knowing that fixing this will require a rewrite of many class methods that expect strings instead of objects.

Okay so if you want to do an abstract class, why not just add the methods to the helper class?

Secondly I would like to insure that automated concepts that are dependent on the string implementation are not lost in this transition. So that mean lets take these steps slower and more patiently and not so fast that we break things that now work.

So I suggest a branch just for this, and we can slowly make this move without effecting the continued development of minor improvements.

@stutteringp0et if you then make pull request for this please target the **simplexml** branch, I will push it up after this update is released.

ro-ot: Sure I can also help with this, I can do the:
> hard work is to identify all the areas where this adaptation will impact the logic of the original string manipulation implementation

So I will start working on that.

ro-ot:
Upgrading from 2.6.5 I get this error:
image

Llewellynvdm:
happy new year!!!

Issue #198, opened 2017-12-12
JText - other than the "_" method


stutteringp0et:
This is more of a note than an issue - because the current extension does function in the prescribed manner - but it is missing a few pieces that I am accustomed to using.

I attempted to use JText::script('...') and found that it totally didn't work.  If you haven't used it, the functionality inserts a translatable string into a javascript var in the output so your javascript can utilize translations via the Joomla.JText._('...') function.  For strings that are only used in Javascript output, there isn't a facility to add these translations.

Any feedback would be welcome.

Llewellynvdm:
Hmmm yes at this time I have only targeted the two php options:
```
JText::_()
JText::sprintf()
```
You can see it done here on line [3110](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler/a_Get.php#L3110)

I suppose we can add the JavaScript equivalent. I have not used it, I saw it in some of the core but I have not paid much attention to it at the time, as you can see over here from line [192](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/views/get_snippets/view.html.php#L192)

So give me some more info and we can add this options...

stutteringp0et:
JText::script(...) works a lot like JText::_(...), except the output isn't to the page where the JText method exists, it becomes part of a system JDocument::scriptOptions(...) call where all strings called by JText::script(...) are added to a javascript object delivered in the document head.  Pretty much any J3.x page will contain a <script type="application/json"... and that object will contain an attribute named joomla.jtext which is used by the window.Joomla.JText._ method to deliver translated strings to javascript on the page.

For the time being, I've been I've been assigning the strings to an unused array so they get assigned constants by JCB, and then later I re-write the script to use the constants in a JText::script(...) statement and rewrite the Javascript to use the correct constants.  It's terribly inefficient.

stutteringp0et:
JText::script(...) works a lot like JText::_(...), except the output isn't to the page where the JText method exists, it becomes part of a system JDocument::scriptOptions(...) call where all strings called by JText::script(...) are added to a javascript object delivered in the document head.  Pretty much any J3.x page will contain a <script type="application/json"... and that object will contain an attribute named joomla.jtext which is used by the window.Joomla.JText._ method to deliver translated strings to javascript on the page.

[...]

PR #199, opened 2017-12-13
moving js out of get_snippets view.html.php

stutteringp0et:
Pull Request for Issue #198 gh-198 .

### Summary of Changes

moved javascript from php to separate js file
changed js string injection to use the Joomla provided method (JText::script())
modified JS to utilize the Joomla JS translated string method. (Joomla.JText._())

### Testing Instructions

use get_snippets

### Expected result

it works

### Actual result

it works

none


Llewellynvdm:
I must tell you (this really helps) I can now clearly see what I should do to make this work in JCB and add it to the compiler.

But okay we have a few issues here, as you may know JCB is build with JCB right... do you understand what that means? Some parts of JCB is build in JCB.

Here is a few screen shots:
image
image
image

So that means it is only the compiler that we can directly edit but the views, the models, controllers are all edited via JCB itself.

### What that means at this time:
When it comes to interface changes I must do it in the JCB interface, buy the Dev's that have the JCB mapped in JCB. At this time only @ro-ot and I have this mapped JCB in JCB package, when I make changes to JCB I share it with him via a Dropbox file we share, same if he makes changes I get a update, so each of us just imports it again and therefore have the latest version and changes in his JCB. At this time I am the only one pushing the changes to the repository we make to the Dropbox package.

That is why in the [issue](https://github.com/vdm-io/Joomla-Component-Builder/issues/198) I asked for examples so I can do that change, sorry if I did not make that clear enough. Here I made a [quick little tutorial](https://youtu.be/4pNBMOYoJpI) to try and explain the work flow a little more. May be I should add it to the pull request guidelines so all understand how it works. There is so much happening around JCB that I am little behind in setting all things in place to insure all understand, but I am doing my best.

Your eagerness to contribute is truly appreciated, and I hope that the little tutorial and this bit of explanation will help you navigate the poetry :+1: with more ease.

Llewellynvdm:
Okay I just added the tutorial to the [contribution guidelines](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/.github/CONTRIBUTING.md)

Please give my your feedback on this if it is clear and really what you think... I am very interested what other in the community think :+1:

stutteringp0et:
I understood that JCB is built with JCB.  That's the main reason I didn't populate the script translations using a loop.  I tried to create what I thought the compiler would create once JText::script was added and the javascript was removed to a different part of the view form (which is also why I placed the javascript in the assets folder instead of where I prefer in a media subfolder.)

You asked for an example - so I gave one that I thought would be informative.

Llewellynvdm:
Tank you very much, it was very informative and will pave the way to improve JCB!!!

Yet looking at the work flow and what I explained as the road map forward, what is your take on that :)

stutteringp0et:
It is enlightening to see how JCB is organized within JCB - that's got to be massive.  how many thousand fields?

stutteringp0et:
I'm going to take the conversation back to the issue - there's the question of the need for known constants in javascript.

Llewellynvdm:
Yes it is massive, but JCB is by far not the largest component I have build with JCB. 

JCB has **1529** fields at this time :+1: I just added a field counter to JCB so we can see that per component after compilation... lol

Okay yes we can continue the conversation on the issue, I will do some changes in the compiler first and then push them up and see if I understood.

PR #220, opened 2018-01-17
Easy Translation via excel

Llewellynvdm:
Added easy translation export and import feature, to allow translation to be done via an excel spreadsheet.

### Summary of Changes
Improved the export and import of Language Translations

### Testing Instructions
- Create more languages 
- Export Language Translations
- Add translation for the new languages
- Import the changed excel with the new translation strings

### Expected result
Will via excel be able to add more translations.

### Actual result
Was not able to do this before.

Will need to make some tutorials to demonstrate this new feature.


admin/helpers/compiler/e_Interpretation.php
```diff
@@ -8104,32 +8104,36 @@ public function setModelExportMethod($viewName_single, $viewName_list)
...
-			$query .= PHP_EOL . "\t\t\t\$headers = new stdClass();";
-			$query .= PHP_EOL . "\t\t\tforeach (\$columns as \$column => \$type)";
-			$query .= PHP_EOL . "\t\t\t{";
-			$query .= PHP_EOL . "\t\t\t\t\$headers->{\$column} = \$column;";
-			$query .= PHP_EOL . "\t\t\t}";
-			$query .= PHP_EOL . "\t\t\treturn \$headers;";
-			$query .= PHP_EOL . "\t\t}";
-			$query .= PHP_EOL . "\t\treturn false;";
-			$query .= PHP_EOL . "\t}";
+			// set some placeholders just incase
```
ro-ot: Just wondering why the placeholders need to be set, it is suppose to be available here. Where is it being dropped?

admin/helpers/compiler/e_Interpretation.php
```diff
@@ -8104,32 +8104,36 @@ public function setModelExportMethod($viewName_single, $viewName_list)
...
+			{
+				$this->placeholders['###views###'] = $viewName_list;
+				$this->placeholders['[[[views]]]'] = $viewName_list;
+				$_viewsSet = true;
+			}
+
+			// add getExImPortHeaders
+			$query .= $this->getCustomScriptBuilder('php_import_headers', 'import_' . $viewName_list, PHP_EOL . PHP_EOL, null, true, 
+				// set a default script for those with no custom script
+				PHP_EOL . PHP_EOL . $this->setPlaceholders(ComponentbuilderHelper::getImportScripts('headers'), $this->placeholders));
```
ro-ot: I like :+1:

site/helpers/componentbuilder.php
```diff
@@ -723,7 +723,7 @@ public static function typeField($type,$option = 'default')
 			
 	public static function getImportScripts($type, $fieldName = false)
```
ro-ot: Is this script really needed in the front-end?

ro-ot:
@Llewellynvdm we should also test it with language strings that has some JavaScript and html in it.

site/helpers/componentbuilder.php
Llewellynvdm: Sure we can move it to only the back-end not big deal.

admin/helpers/compiler/e_Interpretation.php
Llewellynvdm: ah found it, this must have been why so many things lost dynamic `[[[view]]]` replacement.

The `setFieldSet()` method in the `Fields` class cleared it out on [line 757](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler/c_Fields.php#L757) not sure why I did that, besides that the list view was not set at that time and was often needed. But okay I have adapter the compiler to address this issue. Let me know how it works now.

PR #228, opened 2018-01-31
External Code Inclution Option

Llewellynvdm:
### Summary of Changes
Added the feature to include code from external source with the use of a placeholder tag.

### Testing Instructions
Create some code snippet on [Gist](https://gist.github.com) here on [Github](https://github.com), then get the url to the [raw string](https://gist.githubusercontent.com/Llewellynvdm/e13ed45e73590cc4bcd98a73cf9305f1/raw/fancyDate.txt) of the code snippet. Then add the code to a custom area anywhere in your component. Using the following placeholder:
```
[EXTERNALCODE=https://gist.githubusercontent.com/Llewellynvdm/e13ed45e73590cc4bcd98a73cf9305f1/raw/fancyDate.txt]
```
So the placeholder format is
```
[EXTERNALCODE=https://theUrlToText.txt]
```

### Expected result
On compilation the compiler will get the code/string from `https://gist.githubusercontent.com/Llewellynvdm/e13ed45e73590cc4bcd98a73cf9305f1/raw/fancyDate.txt` and add it to the area the placeholder is found.

When a change in the code/string is detected a notice will be given.

### Actual result
New feature!

Will still make tutorials to announce this feature.


admin/helpers/compiler/a_Get.php
```diff
@@ -3651,19 +3658,95 @@ public function setDynamicValues($string)
...
+				// build local bucket
+				foreach ($found as $url)
+				{
+					$key = '[ONLIN' . 'ECODE=' . $url . ']';
+					$urlKey = trim($url);
+					// check if we already fethced this
+					if (!isset($this->onlineCodeData[$urlKey]))
+					{
+						// get the data string (code)
+						$this->onlineCodeData[$urlKey] = ComponentbuilderHelper::getFileContents($urlKey);
```
Rolinevdm: I would add some more validation here?

admin/helpers/compiler/a_Get.php
```diff
@@ -3651,19 +3658,95 @@ public function setDynamicValues($string)
...
+			// urls content
+			$bucket = array();
+			$found = ComponentbuilderHelper::getAllBetween($string, '[ONLIN' . 'ECODE=', ']');
+			if (ComponentbuilderHelper::checkArray($found))
+			{
+				// build local bucket
+				foreach ($found as $url)
+				{
+					$key = '[ONLIN' . 'ECODE=' . $url . ']';
+					$urlKey = trim($url);
```
Rolinevdm: Should we not check that this is an url, or path?

ro-ot:
Should this not instead be called `EXTERNALCODE` since paths also work. I know it is little late in the game, but it makes more sense to call it EXTERNAL.

Llewellynvdm:
Well if we want to change it, best do it now. Please just note that I am not adding a DB fix, you that have been testing this, will have to fix your own DB's to have the correct table and column naming. I will update it for the release.

I will need you all to run the test again, since many variables will change, just to be save 👍

Llewellynvdm:
Are we ready to go?

Issue #230, opened 2018-02-15
Adding SSH


Llewellynvdm:
I have been working on a feature to add ssh to JCB so we can not only FTP files from the developing environment to other locations, but also via SSH move our zipped files around.

So FTP is now being changed to SERVER, and we are adding a protocol type to the view with ssh as a new option.

IF any of you have knowledge on the subject, then please join me on the implementation of this new feature.

Llewellynvdm:
I am planning to make use of the [phpseclib](https://github.com/phpseclib/phpseclib) to achieve the most secure, easy way to do this.

If any of you have a better idea please speak up. Reality is making composer part of JCB would add new and great possibilities.

ro-ot:
I think that phpseclib is the best choice, since it is a pure PHP library.

I would think that you should not include the Composer json and lock file, as this is a limitation for those of use who may want to add other libraries. During update/upgrade of JCB you will mess with our install, by over writing our json and lock file.

Llewellynvdm:
I was wonder over this, I will take it out. But you will have to have the phpseclib as part of your initializer file.

I am working on the implementation of the ssh into JCB as we speak.

PR #232, opened 2018-02-18
Adding SFTP protocol

Llewellynvdm:
Pull Request for Issue gh-230.

### Summary of Changes
Adding SFTP, so we can via SFTP push updates and sales server files.

### Testing Instructions
Add a SFTP settings to the new server area and see how it works.

### Expected result
The package will be moved to the server via SFTP

### Actual result
New Feature

May need to make a new tutorial on how to use this feature, but will first expand on this a little more before making the tutorial.


admin/helpers/componentbuilder.php
```diff
@@ -1932,10 +1932,171 @@ public static function getFileContents($path, $none = '')
...
+
+	/**
+	* 	get the server details
+	* 	
+	* 	@param   int    $serverID       The server local id to use
+	* 	@param   int    $protocol       The server protocol to use
+	* 	
+	* 	@return  array    on success with sftp server details
+	**/
+	protected static function getServerDetails($serverID, $protocol = 2)
```
Llewellynvdm: I think more permission checks are needed here, so we can have little more control over this. Powerful feature, but with power comes danger if not control is set in place.

Issue #265, opened 2018-04-15
Suggestion for naming consistency on Import Components/Import JCB Packages


kevincam3:
On the dashboard you see the option "Import JCB Packages"
![image](https://user-images.githubusercontent.com/870934/38782772-b967bcb0-40ce-11e8-967f-89d6747e1968.png)

On the Joomla Components page you see the option "Import Components"
![image](https://user-images.githubusercontent.com/870934/38782781-dce95efa-40ce-11e8-8c25-267d6e99d60d.png)

I was wondering if they do different things but appear to take you to the same page so I was thinking they should be called the same thing.

Issue #261, opened 2018-04-12
Dynamic pull translations from Joomla where they already have those strings translated


bosyrih:
### Feature Request (new issue)
Can we make the translation language take language translation from joomla translation and concat it with component name, with this method we can save a lot of time with translation.
you know that joomla had translated to 74 language that mean all of main words and terms have translate like (save,edit,permission ...) .
so we can find the language the JCB installed in joomla and take the translated string and concat them with component name

### Steps to reproduce the issue (original issue - resolved)
1- go to language. define new languages , name: Arabic, language tag: ar-AA with space after AA.
2- go to 	Language Translations and translate some text,
3- compile component. meesage about languge installer that can't be copy after installation.
4- go back to step 1 and correct language tag to be ar-AA without space after AA.
5- go to 	Language Translations and translate some test.
6- compile component. there is two language file on with ar-AA without space and one with space.
and you need to resave all the text with ar-AA with space with the correct language tag.

### Expected result
when you change language tag from languages your old 	Language Translations text that connected with this language name must be change too.

### Actual result
lose of all language tag in Language Translations texts. and you need to resave them.

### System information (as much as possible)
- OS Name & Version: 
- MySql Version: 
- Apache Version:  Apache/2.4.29 (Win32) OpenSSL/1.1.0g PHP/7.2.2
- PHP Version: 7.2.2
- Joomla Version: 3.8.5
- JCB Version: 2.7.1
- Browser:

Llewellynvdm:
The best I can do is to trim empty space around the language tag, to insure it does not happen again, that a typo can cause so much pain. But for the need of fast execution during compilation we link the strings to the actual language tag and not the id. This can not be changed, sorry!

Llewellynvdm:
So in the next update the patch will be included to fix this, let me know if you have any further questions.

bosyrih:
Hi,
thanks .
Can we make the translation language take language translation from joomla translation and concat it with component name, with this method we can save a lot of time with translation.
you know that  joomla had translated  to 74 language that mean all of main words and terms have translate like (save,edit,permission ...) .
so we can find the language the JCB installed in joomla and take the translated string and concat them with component name

Llewellynvdm:
This is a great idea, I will give it some thought and see if we can automate this, so as to grab the translated values from Joomla and dynamical translate all values that exist in the Joomla translation.

I will add this to the feature request!

Llewellynvdm:
> How I think it should be done

[...]

PR #270, opened 2018-04-26
Adding Expansion Development Method and more...

Llewellynvdm:
Pull Request for many Issue

### Summary of Changes
- Added a switch to control the message display of the expand feature when cronjob runs
- Added the feature that allows you to setup the behaviour of a selected set of components that will be auto build and installed. So you do not need to manually compile and install those components any more, the system does all that automatically for you. We have called the feature Expansion Development Method, and it can be set in the global settings of JCB under the Development Method tab.
- Fixed the auto backup system, to allow JCB to be backed up automatically
- Tweaked the field database area to insure that users are forced to set the database values
- Fixed a strange mismatch issue with JCB package import with more then 10 components at a time. Also fixed the translation entranslation to source.
- Improved JCB to now be able to have other languages then en-GB as its base/main/global language. This will allow you to build your components in a language of your choice.
- Added edit links to templates and layouts.
- Fixed the alias and title selection issue in admin fields area.
- Added all available languages in Joomla to language area.
- Added minify of JS to the compiler area, also improved the infusion area to make more use of the getCustomScriptBuilder method
- Fixed the Add JavaScript (views-footer) option, since it did not add the javascript, but now does. Also added some spinner to the Joomla Components view.
- Added the option to clone a Component and all its linked data, also to import a JCB package without merging with local Component
- Keeping naming consistency resolved gh-265. Started adding the needed features for the new translation improvement gh-261
- Added the php field properties to a textarea outside of the subform for easier editing

### Testing Instructions
- Check that the demo component is able to compile and that you can all over JCB edit, create and all the other normal functionality still works as expected.
- Check that you can import the demo component without merging, and so have duplicates of all fields, views and other linked data to the component.
-  Test drive JCB with other language as en-GB being the main language.
- Check that the auto back-up system still works.
- Test the Expansion Development Method, it can be set in the global settings of JCB under the Development Method tab.

### Expected result
- That JCB is stable and now having more features and improvements then before.


admin/helpers/componentbuilder.php
```diff
@@ -2256,6 +2290,607 @@ public static function composerAutoload()
...
+	/**
+	*	To Lock string
+	*
+	*	@param string  $string       The string/array to lock
+	*	@param string  $key          The custom key to use
+	*	@param int      $salt           The switch to add salt and type of salt
+	*	@param int      $dynamic    The dynamic replacement array of salt build string
+	*	@param int      $urlencode  The switch to control url encoding
+	**/
+	public static function lock($string, $key = null, $salt = 2, $dynamic = null, $urlencode = true)
```
ro-ot: This is cool, thanks for sharing!

admin/helpers/componentbuilder.php
```diff
@@ -2256,6 +2290,607 @@ public static function composerAutoload()
...
+	*
+	* 	@param  string      $url               The url of where the task is to be performed
+	*  	@param  array       $_options      The array of curl options/headers to set
+	*	@param  function   $callback      The option to do a call back when task is completed
+	*	@param  int           $threadSize   The size of the thread
+	*
+	* 	@return  bool true   On success
+	*
+	*/
+	public static function curlMultiExec(&$url, &$_options, $callback = null, $threadSize = 20)
```
ro-ot: I like :+1: almost multithreading in PHP wow!

ro-ot:
@Llewellynvdm so far it looks good, I will give you final feedback over the weekend.

admin/helpers/componentbuilder.php
Llewellynvdm: Pleasure! simple little trick to lock values for a short period of time.

ro-ot:
I think best let this release go, since there are so many nice fixes in it, and make that feature the next release. But it is really up to you.

Issue #287, opened 2018-05-18
Being able to join/combine result of multiple fields in admin list view.


Llewellynvdm:
### Feature Request
The request came from the JAB to be able to combine/join multiple fields into one for display in the list view of the admin area.

This option so be able to add calculations, or even model the values into one result.

mwweb:
That would be a cool feature. I currently did this with custom code, but would be nice to have as part of the core.

PR #293, opened 2018-05-26
Added the Feature to Combine/Join Values/Fields in the admin list view (called field relations)

Llewellynvdm:
Pull Request for Issue gh-287.

### Summary of Changes
- Resolved gh-287 that adds the feature to join/combine results of multiple fields in the admin list view.
- Update the tab string to make tabs dynamic in nature.
- Improved the compiler building of the admin list view
- Refactored the compiler list view body builder, in preparation of the relation fields.
- Added more options to the relation admin fields. gh-287
- Refactored the compiler update file method
- Updated compiler to allow dynamic update for language strings. Started adding the field relationship feature requested in gh-287

### Testing Instructions
Watch [the tutorial and do a combine/join](https://youtu.be/hh4IIPmYIY8) fields in the admin view.

### Expected result
The result demonstrated in the tutorials.

### Actual result
New feature

The new tutorial will be added to the JCB playlist on youtube once the feature is released.


mwweb:
I tested this, and it appears to work. I will say that this appears to be a nice feature, but honestly don't see where I'll end up utilizing it in any of my components, at this point.

I have a similar feature right now, all using custom code, that takes what you did here a step further. I'll "throw this out" as something to think about.

In my case, I have a field called "Name". On save, if name is populated, then nothing gets done. However, if Name is empty, then it concatenates fields. Those concatenated fields are then saved in the Name field in the DB.

In this case, the name field I just called "Adline" which concatenates a unique record id, street, city, state, and postal code, if there is nothing manually typed in Adline.  So, if the unique record id is a12345, street=123 Front Street, city=New York, state=NY, postal code=10001, then on save name becomes a12345 123 Front Street New York, NY 10001. I can then have that field, along with the street, city, state, etc on the list view.

That night be something to think about for the future of extending this functionality.

Llewellynvdm:
@mwweb I don't understand, this is what this new feature can do. Well basically the same and so much more. The ability to combine values in the list view.

The feature you are speaking about also is in JCB where you have fields in the admin edit view that on save gets combined into the database as one column. 

But this feature we are looking at takes values from different columns and combines them in display only, while still allow the filtering and searching of those columns values separately.

I knew that with custom scripting you can already do this, but it seemed to me from reading the issues on github and feedback from the JAB that this needed to be done easier.

Okay so in my mind it seems like the header of the table is the issue here.
untitled

So to have the option to change the word **Name** should not be that hard... I can just add another field to override the header.

When I opened the issue gh-287 you were all for this feature. Now that I spend a week to build it you are like... ahaa wast of time! Facts are the feature does exactly what it was said to do, it combines/joins values in the **admin list view** in a very powerful way, and very easy.

If the **greatest motivation** for the feature is the limited space why add city again, or any other value that is being combined. Truth is the filter can still be loaded for those fields, and so also the search options. Not being able to add the field/value a second time as a **list field** is never going to change (sorry to say), but you can use any of the combine/**join fields** again and again... so there is the loop to use one field more then once in the list view.

Lastly, you do know that we added the option to build the alias dynamical with that kind of way you explained? well to do that with every field will be a complete other feature. Not the same as this one, as this feature focuses on the display of the value and not the creation/storing of the values. Further more as you said you already have nice custom script to deal with that. What is also true, is that you can achieve the same effect (in display) with this feature.

lamatieregrise:
Nice Feature guys!

admin/helpers/compiler.php
```diff
@@ -72,7 +71,7 @@ public function __construct($config = array())
 				// first remove the files and folders
 				$this->removeFolder($this->componentPath . '/site');
 				// clear form component xml
-				$xmlPath = $this->componentPath . '/' . $this->fileContentStatic['###component###'] . '.xml';
+				$xmlPath = $this->componentPath . '/' . $this->fileContentStatic[$this->hhh . 'component' . $this->hhh] . '.xml';
```
ro-ot: This seems promising!

admin/helpers/compiler/c_Fields.php
```diff
@@ -457,7 +457,7 @@ protected function stringFieldSet($view, $component, $view_name_single, $view_na
 		$readOnly = false;
 		if ($view['settings']->type == 2)
 		{
-			$readOnly = "\t\t\t" . 'readonly="true"' . PHP_EOL . "\t\t\t" . 'disabled="true"';
+			$readOnly = $this->_t(3) . 'readonly="true"' . PHP_EOL . $this->_t(3) . 'disabled="true"';
```
ro-ot: This is a great move. We should also do this for the template files, or something like this.

Llewellynvdm:
The JavaScript should be more stable now, and also added the option to over-ride the column header via the relation area.

Issue #325, opened 2018-07-23
Button 'Import JCB Packages' on control pannel gives error


Schrijvers123:
### Steps to reproduce the issue
Click on main menu on component > component builder.
CLick on the button 'Import JCB Packages' 

### Expected result
The import screen

### Actual result
Error with the message: 'The most recent request was denied because it had an invalid security token. Please refresh the page and try again.'

Via the Button Joomla Component and then i the top menu Import JCB Package it is working.

### System information (as much as possible)
- OS Name & Version: Windows 10
- MySql Version: 5.5.5-10.1.34-MariaDB-1~xenial
- Apache Version: 
- PHP Version: 7.0.30
- Joomla Version:  3.8.10
- JCB Version: 2.8.4 staging
- Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:61.0) Gecko/20100101 Firefox/61.0

Llewellynvdm:
hmm security token, this must have been because I added a token validation to the smartImport method in the controller. Thanks I missed that.

Schrijvers123:
Test the new version and it is working.

Issue #324, opened 2018-07-22
FieldTyep CusomUser is save wrong


Schrijvers123:
### Steps to reproduce the issue
Make a new CustomUser field. And save it.

### Expected result
After saving the new field, the php code in getExcluded methode should be normal as normal. The default php code or code that you changed by yourself.

### Actual result
The php code that is used in the getGroups methode.

### System information (as much as possible)
- OS Name & Version: Windows 10
- MySql Version: 5.5.5-10.1.34-MariaDB-1~xenial
- Apache Version: 
- PHP Version: 7.0.30
- Joomla Version: Joomla! 3.8.10 Stable [ Amani ] 26-June-2018 15:45 GMT
- JCB Version: 2.8.4 staging
- Browser: Firefox (latest)

Llewellynvdm:
Okay it has been fixed in the last [push to the staging branch](https://github.com/vdm-io/Joomla-Component-Builder/commit/812857da813d5863a8043b2b73e0fdc2fd780127#diff-78b2a246848801b8f5bb5e1521037903L930).

Issue #322, opened 2018-07-20
default field ID is writeble during editing component


Schrijvers123:
### Steps to reproduce the issue
After build your component an goto a single admin view. Click on the tab Publish, and look the field ID. 

I expected that it is read only but actual it is an edible field.

### System information (as much as possible)
- OS Name & Version: Windows 10
- MySql Version: 5.5.5-10.1.34-MariaDB-1~xenial
- Apache Version: 
- PHP Version: 7.1.19
- Joomla Version: 3.8.10
- JCB Version: 2.8.4 (staging)
- Browser: Firefox

### Additional comments
On the place where the id xml field is added, it needt to have an extra parameter readonly="true"

Llewellynvdm:
There is, when you create any field, and you select the field type, and it is missing a property, then just edit the field_type and add the property. Just make sure that it is a valid property in the Joomla xml field implementation of that field_type.

So you can click on the small blue pencil icon to edit the field_type:
![image](https://user-images.githubusercontent.com/5607939/43025968-a1008c54-8c6b-11e8-9993-f05424ae818d.png)
Or you can go to field types and just edit the field_type in question. Please[ watch this tutorial again](https://youtu.be/OhLzvThDXls?t=9m39s) if you need more help with this.

Schrijvers123:
I know you could change it, but an id fiels should be by default read only. Or am I missing something?

Schrijvers123:
Thanks for the fast response and explanation. I will check and i make sense.

Also I will look into the code of JCB to get familiar with it so I can help when necessary. I'm realy impressed what you made.

lnk0bel:
I did not override the id field on my side and it is not read only either so it seems it's no more read only by default even when not overridden.

Created by, modified by, modified date and revision are read only but not the id.

Llewellynvdm:
O yes it is, [here you can check the code](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler/c_Fields.php#L476) in the compiler that manages the ID. O wait a moment it is the xml implementation that is not adding the read only [over here](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler/c_Fields.php#L744). 

[...]

Issue #318, opened 2018-07-12
Add standard/default fields to list view tries to duplicate field


cpaschen:
### Steps to reproduce the issue

Following the video tutorial here:
https://www.youtube.com/watch?v=FHQfIhWHYyQ&feature=youtu.be&list=PLQRGFI8XZ_wtGvPQZWBfDzzlERLQgpMRE&t=5m45s

Add a field element for an existing default field (like 'created_by').
Add to admin view with display/filter/etc as desired

### Expected result

Should add to the list view (and filters, and to different tab than publishing if desired) WITHOUT trying to add a duplicate field to the table.

### Actual result
Gives error: "Duplicate column name 'created_by' " and an error installing component.

### System information (as much as possible)
- OS Name & Version:  WAMP
- MySql Version: 5.5.56
- Apache Version: 2.4.27
- PHP Version: 7.1.9
- Joomla Version: 3.8.8
- JCB Version: 2.8.0
- Browser: multiple

### Additional comments

Work-around per forum does resolve the issue.

(remove the SQL string in the component updates that is trying to create the field AND remove the /administrator/COMPONENT/sql/updates/mysql/x.x.x.sql file where that update is being added as well, then compile install)

PR #326, opened 2018-07-30
Few Fixes and Improved the Admin Fields Relations

Llewellynvdm:
Pull Request for Issues gh-325, gh-324, gh-322, gh-318, gh-317 .

### Summary of Changes
- Fixed gh-325 that caused a security token validation error.
- Fixed gh-324 that had a typo with the storing of the PHPx values.
- Improved the safeString helper method
- Improved the Admin Fields Relations area to allow more custom coding without the need of selecting Join Field/s.
- Improved the dynamic get string fixing implementation to avoid conflict when the same field name is used in multiple joined tables requiring string fixing.
- Fixed gh-322 to insure ID is always readonly even if override detected.
- Added new feature to control access and view/hide control per/field.
- Fixed gh-318 so that default and other none database fields will not trigger SQL update and version increment.
- Fixed gh-317 that had conflict variable naming in the jsonToString helper method. Improved the refresh method for JCB package area.

### Testing Instructions
Please read all the fixed issues and do the test surrounding those areas. Most of them have already been tested, and confirmed fixed inside the given issues.

### Expected result
That JCB is more stable then before, with a few new features.

### Actual result
We had some bugs (which has been fixed), and new features added.

We may need some tutorial on the new permission feature, just so that it gets introduced, and some update to the Admin Fields Relations area.


admin/helpers/compiler/a_Get.php
```diff
@@ -1540,112 +1547,123 @@ public function getAdminViewData($id)
...
-									}
-									elseif ($old_field_name !== $field_name)
-									{
-										// give a notice atleast that the multi fields could have changed and no DB update was done
-										$this->app->enqueueMessage(JText::sprintf('You have a field called <b>%s</b> that has been added multiple times to the <b>%s</b> view, the name of that field has changed to <b>%s</b>. Normaly we would automaticly add the update SQL to your component, but with multiple fields this does not work automaticly since it could be that noting changed and it just seems like it did. Therefore you will have to do this manualy if it actualy did change!', $field_name, $name_single, $old_field_name), 'Notice');
-									}
-									// remove tmp
-									unset($tmpfield);
+									// this only works when the field is not multiple of the same field
+									$this->setUpdateSQL($old_field_name, $field_name, 'field.name', $name_single . '.' . $field_name);
```
ro-ot: If you change a none-database field to default it does not get detected.

admin/helpers/compiler/c_Fields.php
```diff
@@ -749,6 +742,7 @@ protected function simpleXMLFieldSet($view, $component, $view_name_single, $view
 				'name' => 'id',
 				'type' => 'text',
 				'class' => 'readonly',
+				'readonly' => "true",
```
ro-ot: Oops @stutteringp0et

Llewellynvdm: Well we also over looked it :mag:

admin/helpers/compiler/a_Get.php
Llewellynvdm: Well for now the huge warning when making a field none-database is the best we can do. Since to catch this will be way to much... just for one out of a million. So lets keep a look out... and if it comes up in the forums we can look at it again.

Issue #349, opened 2018-10-04
DynamicGet | Main Source: Custom 


Schrijvers123:
### Steps to reproduce the issue
Make a new dynamiGet with main source as Custom.

In the CustomGet place the code:
```
$query
			->select($db->quoteName(
				array('singletraining.id','singletraining.trainingroom','singletraining.startdate','singletraining.enddate','singletraining.trainingname'),
				array('id','trainingroom','startdate','enddate','trainingname')))
			->from($db->quoteName('#__trainershub_singletraining', 'singletraining'))
			
			->select($db->quoteName(
				array('location.address','location.city','location.country','location.locationname','location.postalcode'),
				array('trainershub_location_address','trainershub_location_city','trainershub_location_country','trainershub_location_locationname','trainershub_location_postalcode')))
			->join('INNER', ($db->quoteName('#__trainershub_location', 'location')) . ' ON (' . $db->quoteName('singletraining.location') . ' = ' . $db->quoteName('location.id') . ')')

		
			->select($db->quoteName(
				array('customer.companyname'),
				array('trainershub_customer_companyname')))
			->join('INNER', ($db->quoteName('#__trainershub_customer', 'customer')) . ' ON (' . $db->quoteName('singletraining.customer') . ' = ' . $db->quoteName('customer.id') . ')')
		
			->where('singletraining.status != 3')
			->where('singletraining.private_training = 0')
			->where('singletraining.published = 1')
			->where('yearweek( singletraining.startdate ) = yearweek( curdate( ) )')
			->order('singletraining.startdate ASC');

		// return the query object
		return $query;
```
Compile the code and see in the modell that the code is different:

```
protected function getListQuery()
	{
		// Get the current user for authorisation checks
		$this->user = JFactory::getUser();
		$this->userId = $this->user->get('id');
		$this->guest = $this->user->get('guest');
		$this->groups = $this->user->get('groups');
		$this->authorisedGroups = $this->user->getAuthorisedGroups();
		$this->levels = $this->user->getAuthorisedViewLevels();
		$this->app = JFactory::getApplication();
		$this->input = $this->app->input;
		$this->initSet = true; 
		// Get a db connection.
		$db = JFactory::getDbo();

		// Create a new query object.
		$query = $db->getQuery(true);

		// Get data
		$query
			->select($db->quoteName(
				array('singletraining.id','singletraining.trainingroom','singletraining.startdate','singletraining.enddate','singletraining.trainingname'),
[...]

Issue #345, opened 2018-09-23
Feature Request: Select mysql table engine.


peterpetrov:
This feature request is to be able to select the engine for table creation when creating new Admin View. Currently it is only MyISAM. The benefit will be to create INNODB which will be better for some tables when the website is under heavy usage. The main difference is that MyISAM locks the whole table to insert or update a record. INNODB locks just the line in question. That can significantly decrease the load time for sites with a lot of content built with JCB.

Llewellynvdm:
Okay in the admin view under the mySql tab there are three new select options, to tweak the table, give it a try and let me know it you run into any trouble.

This was a big changes, and I have tried my best to have it work in all shapes and sizes out of the box... hope that will be the case.

peterpetrov:
It works perfectly. Thank you very much for that.

Llewellynvdm:
Tell me @peterpetrov why would this error occur as @mwweb described in gh-348

> 
> JInstaller: :Install: Error SQL Row size too large (> 8126). Changing some columns to TEXT or BLOB or using ROW_FORMAT=DYNAMIC or ROW_FORMAT=COMPRESSED may help. In current row format, BLOB prefix of 768 bytes is stored inline.
> Extension Update: SQL error processing query: DB function failed with error number 1118
> Row size too large (> 8126). Changing some columns to TEXT or BLOB or using ROW_FORMAT=DYNAMIC or ROW_FORMAT=COMPRESSED may help. In current row format, BLOB prefix of 768 bytes is stored inline.
> SQL =
> ALTER TABLE #__componentbuilder_joomla_component ENGINE = InnoDB;

I think it is a local limitation, could I be wrong...

peterpetrov:
That has to be local one. I will comment on the issue itself.

Llewellynvdm:
Yes I did move the whole of JCB to InnoDB, that is the issue...

The file that does this is [2.9.5.sql](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/sql/updates/mysql/2.9.5.sql)

Issue #341, opened 2018-09-17
Feature request: Extending custom fields to allow getinput override


AlexDings:
### Expected result
As custom fields are used within JCB to allow for field-type changes, I would like to be able to overwrite the getInput() function of a custom field, the same way it is currently possible to do this for the getOptions() function.

This would allow people to have more control over the field's appearance and behavior.
An example would be to easily add extra data parameters to the field.
Referencing this: [Joomla documentation: Creating a custom form field type](https://www.google.com/url?q=https%3A%2F%2Fdocs.joomla.org%2FCreating_a_custom_form_field_type&sa=D&sntz=1&usg=AFQjCNFjuhEgDxBIe95Aps-vrmz-m1DxFw)

### Actual result

New feature: getInput() overwrite ability within custom fields in JCB..

### Additional comments

Follow up from this forum post:
[Custom field types](https://groups.google.com/a/vdm.io/forum/?utm_medium=email&utm_source=footer#!topic/jcb/nS24TX0CUik)

Llewellynvdm:
Okay I think I have it ready for testing... it came out little different then I planned originally, but I think you will like how it works.

Here is a [quick tutorial](https://youtu.be/Mu9H0zgH9Lw) on how to use this feature. @AlexDings please test this on your side and let me know if it works as expected.

AlexDings:
Thanks for the quick work.
I am quite busy with work right now but will make sure to get you all my feedback by the end of this week

Llewellynvdm:
Lets just say I also needed this feature :+1:

AlexDings:
The feature is looking good, and I could work with it.
However, right now the type_phpa-z values are treated as default values and are overwritten when creating the field which uses this fieldtype.

Would it be possible to make it so that the php for this fieldtype is set, and this can't be changed on the field creation. This would work better for me since I will have 10 fields using this unique list-getinput method. Having this in the fieldtype and not the field would allow me to update it on one place and have it have effect on all the fields using this fieldtype.

Llewellynvdm:
I am not sure I understand, but for what its worth... ones you create the new custom-custom field-type just code what you want, I mean all you want it to do in to the type_php_a and b and c and so on... then when creating the field, just don't edit it, let the default be the actual. Or am I missing something?

You don't have to edit those fields, you can with placeholders make it dynamic like the original custom field.

Llewellynvdm:
[...]

Issue #348, opened 2018-09-27
Feature Request -- Dynamic Get -- Ordering


mwweb:
Currently in the dynamic gets there is the ordering section under the Tweak tab. In most cases, this is perfectly fine. However, let's assume that in Sermon Distributor you have 25 preachers. If you can only sort by ascending or descending, you could be perceived as showing favoritism by only showing the same preachers all the time.

It would be nice to have an option, whether in the Ordering Direction listbox, or a yes/no, or something, called Random. If Random, then instead of `$query->order('a.ordering ASC');`, it would generate `$query->order('RAND ()');`. In doing that, the records on that view would be randomized.

### System information (as much as possible)
- OS Name & Version: 14.04.1-Ubuntu
- MySql Version:  5.5.5-10.1.28-MariaDB-1~trusty
- Apache Version:  2.4.29
- PHP Version: PHP Version 7.1.18-1
- Joomla Version: 3.8.12
- JCB Version: 2.9.0
- Browser: Chrome

Llewellynvdm:
Have a look at this https://stackoverflow.com/a/23706906/1429677

It seems to say that we need a random seed passed to ensure it is random... did I understand this correct. Then reading [this issue](https://github.com/joomla/joomla-cms/issues/16081) on the Joomla repo, it seems Joomla has been looking at making this part of the API.... but I can't find it... can you?

So on [line 1132](https://github.com/joomla/joomla-cms/blob/staging/libraries/joomla/database/query.php#L1132) the order value is loaded, and is triggering the `JDatabaseQueryElement` [class](https://github.com/joomla/joomla-cms/blob/staging/libraries/joomla/database/query/element.php#L21).

The [constructor](https://github.com/joomla/joomla-cms/blob/staging/libraries/joomla/database/query/element.php#L50) loads the `ORDER BY` with `RAND()` in as an element.

So if we want the seed loaded we must do it PHP side, and via JCB we must do something like this:
```
$query->order('RAND(' . rand ( 10 , 500 ) . ')');
```
Will this work?

Llewellynvdm:
I for now just added what you asked... please test and give feedback, also on the first post I made.

mwweb:
In attempting to install that staging branch on a test site:

Warning
[...]

PR #350, opened 2018-10-05
Added InoDB options, Expanded the Custom Field

Llewellynvdm:
Pull Request for Issue gh-341, gh-345, gh-348, gh-349.

### Summary of Changes
- Fixed gh-349 to insure dynamic joined and other tweaks are not added when custom code is selected as main source with a dynamicGet
- We removed the Transition to InnoDB for all JCB tables and moved it to https://gist.github.com/Llewellynvdm/85234c6bd1e7d0eb4ddd6c925770aa2c for those who want to manually make the transition. gh-348
- We also added a pre-flight script to remove all unused columns from JCB.
- Improved the custom-custom field implementation, to allow duplicate fields to be use in sub-forms and in same form/view with the correct expected behavior of incremental naming, only if in same sub-form or form/view. gh-341
- Further fix to the search of customcode placeholders across the system.
- Added the option to random order the dynamic get result set. gh-348
- Fixed the search for customcode placeholders across the system. gh-341
- Changed the text field for Value/Example in the Fieldtype area to a textarea so we can add more code per/line. gh-341
- Little fix to the field builder in the compiler
- Resolved gh-345 by adding whole new mysql table selection options in the admin view, under the mySql tab. Fixed the folder & file loader.
- Fixed the Update of custom files, so even the custom code and external code placeholders will work in external files.
- Resolved gh-341 that adds the feature to expand the custom field option far beyond the default. Moved the menu prefix option in to the Joomla component area -> tab -> Settings to allow each component to override the global setting.
- Added the [Run Expansion] button option for faster compile and install option of selected components.
- Update the Uikit 3 lib. Fixed permission check in getEditUrl.

### Testing Instructions
Please review the various issues gh-341, gh-345, gh-348, gh-349 (most already tested successful)

### Expected result
A stable JCB

### Actual result
Many new features and bug fixes

We have [a new tutorials](https://www.youtube.com/watch?v=Mu9H0zgH9Lw&index=75&list=PLQRGFI8XZ_wtGvPQZWBfDzzlERLQgpMRE) for the new custom-custom field type implementation.


admin/custom/wiki_menu.txt
```diff
@@ -0,0 +1 @@
+[EXTERNALCODE=https://raw.githubusercontent.com/wiki/vdm-io/Joomla-Component-Builder/1.---Home.md]
```
ro-ot: So do I understand this correct, JCB can now update external files with custom code placeholders, as well as external code placeholders. Stunning :1st_place_medal:

admin/helpers/componentbuilder.php
```diff
@@ -2254,6 +2272,38 @@ public static function getPackageComponentsDetailsDisplay(&$info)
...
+			// check if we have a key
+			if (self::checkString($key))
+			{
+				if (strpos($data, $key) !== false)
+				{
+					return base64_decode(str_replace($key, '', $data));
+				}
+			}
+			// fallback to this, not perfect method
+			elseif (base64_encode(base64_decode($data, true)) === $data)
```
ro-ot: This should just be an if, then it will be the default.

admin/custom/wiki_menu.txt
Llewellynvdm: Yes that is a new feature... :+1:

Issue #370, opened 2019-01-26
Feature request - auto update dynamic get when adding fields to admin view


TLWebdesign:
The title says it all. It would be nice when creating a dynamic get that there would be a toggle switch "auto update". If set to yes it would automagically update the dynamicGet with the new field when a new field is added to the admin view. I guess it should only work with getItem and getItemList. not the custom options.

Llewellynvdm:
hmmm lets see, that will mean all the fields are added with the update right?

I can add this, very easy actually. But the drawback is that it will not be selective (all fields will be added), and only work on the **Back-end Views** not the _Joomla Database_ as the source. I can also make it target the **Join View Tables** but again it will not be selective (all fields will be added).

So we will add two switches, one for the **Main Source**, and one for each **Join View Table**. How it will work, at the moment you add the field to the Back-end View (admin view) there will be a function updating all dynamic gets and joints that target that view, if they have the Auto Update selected in the dynamic get. 

Did I understand you correctly? I like the idea, a nice time safer, and these kind of things are always welcome.

ro-ot:
I think sync is a better term here. To keep the selection in sync with the selected view and if that is true then hide the selection text area, or make it read only.

TLWebdesign:
that's nice indeed. I think the toggle switch "auto sync" would be set in dynamic Get. And indeed if set to yes it would just add all fields from the backend view it is connected to. So this could look like this:

![schermafbeelding 2019-01-28 om 12 21 26](https://user-images.githubusercontent.com/4402824/51833270-44ea8f00-22f7-11e9-81b6-b17b3e16b5c4.png)

Llewellynvdm:
Okay I came up with a more elegant solution. Which is more instep with the expected behavior. That is if you too are a coffee drinker :+1: 

Instead of **auto sync** I added a switch to **Select All**, which then adds `a.*` to the selection area and will effectively do just that. This way we don't need extra scripts to update the get, since the query will always retrieve all the fields.

Then to get this to work on the joint areas you simple replace the selection with the befitting `b.*` or `c.*` as related to the AS selection and it will have the same behavior, done!

Take it for a ride and let me know!

TLWebdesign:
Nice solution indeed! 👍 No coffee for me tho hihi 😉

Llewellynvdm:
O yes and it works for moth view and database selection.

TLWebdesign:
Works like a charm! Thanks very much

Issue #369, opened 2019-01-25
Row size too large (> 8126)


Llewellynvdm:
### Steps to reproduce the issue
Import of the JCB package of Joomla Member Manager

### Expected result
That all data just import

### Actual result
We get the following error:
```
Row size too large (> 8126). Changing some columns to TEXT or BLOB or using ROW_FORMAT=DYNAMIC or ROW_FORMAT=COMPRESSED may help. In current row format, BLOB prefix of 768 bytes is stored inline. 
```

### System information (as much as possible)
https://hub.docker.com/r/bitnami/joomla/

### Additional comments
Issue was mentioned on the [JCB forum](https://groups.google.com/a/vdm.io/forum/#!topic/jcb/eFXqITxNFoU)

Llewellynvdm:
Hmm I can confirm when I install JCB v2.9.8 and then check the ROW_FORMAT it is set to be Compact. This seems to be the default for the InnoDB engine (in some environments), as when I change the tables to MyISAM in JCB (and compile and install) then as MyISAM, then they are by default Dynamic (in some environments).

Okay so this does seem like the issue (**ROW_FORMAT**)

We can add this to the admin views just like we did with the engine selection. These are the options I have in mind. https://dev.mysql.com/doc/refman/5.5/en/innodb-row-format.html

I am just wondering if the **file formats**, also needs attention?

TLWebdesign:
From what i read on the interwebs is that setting it to dynamic should fix the issue. Also on our own google group it is given as the fix for another similar issue.

Llewellynvdm:
There is not install.sql in the members manager package. The update to JCB is enough to resolve the issue, as I have tested it a few times with the [bitnami stack deployment](https://hub.docker.com/r/bitnami/joomla) of Joomla in a docker environment.

What I did is I install JCB v2.9.8 on the Bitnami Joomla website, then I used [this link](https://github.com/vdm-io/Joomla-Component-Builder/archive/staging.zip) to update it, en then the same old JMM package on gihub, and that package of 48 component and it all just worked, where it did not before, now it does.

You will see in v2.9.9 we have added an [sql update script](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/sql/updates/mysql/2.9.8.sql) to update all the tables, you should confirm that the tables indeed were updated.

Running that script manually should also work as a fix. Reality is we must convert all the tables to a DYNAMIC row_format, and that script does just that.

[...]

Issue #367, opened 2019-01-22
Feature Request - Add css class option to control-group in edit form


cpaschen:
Add the ability to add a 'wrapper' CSS class to any field added to an admin view so that the class is added to the edit view (admin/front) at the point where the "control-group" class is added to the div.

example of front-end code currently created:

```
<div class="row-fluid form-horizontal-desktop">
  ::before
 <div class="span12">
    <div class="control-group field-spacer">...</div>
    **<div class="control-group">**
      ::before
      <div class="control-label">
        <label id = ....
      </div>
      <div class="controls">
        ...
      </div>
      ::after
   </div>
```

Being able to target this level of div will allow more options in the layout of the edit forms (without needing to manually override the entire edit form layout.

cpaschen:
I realized that adding the ability to add a unique tag to at field via the gui is just asking for more code to maintain (that would probably be less used than most features) and would also increase the business of that already full field form.

Instead, why not just automatically add the field name as a class to each of those control-labels. That already happens at the label level (each label already has a class applied that is unique to that field).
If we just added that same class at the control-label div then it would only take a couple/few lines of code and no gui interruptions.

So the resulting html might look like:

`<div class="control-label  this-field-name">`

TLWebdesign:
That sounds like a good option.

Llewellynvdm:
Due to the nature of complexity here just adding the field name is not that simple, there can be cross matching behavior since Joomla already uses the field name when building the field. Then most importantly, you do Joomla builds the field HTML and not JCB right?

I mean the code that delivers the HTML code you placed in the opening comment of this issue is generated with this line:
```php
echo $form->renderField($f);
```
In some layout template that JCB build... and then Joomla internal classes take over.

[...]

Issue #365, opened 2019-01-13
Custom Code self reference check is off


AlexDings:
### Steps to reproduce the issue
Create a custom code with the name "printSomething"
Type [CUSTOMCODE=printSomethingHelper] inside as if you wanted to include that custom code inside "printSomething".

### Expected result
It saves, as there is no loop or self-reference.

### Actual result
It removes [CUSTOMCODE=printSomethingHelper] and gives the following warning:
~~~~
Warning
Placeholder Removed!
This custom code can only be used in other custom code, not in it self! Since that will cause a infinite loop in the compiler.
~~~~

### Additional comments
In my testing, this happens when the first part of the custom code name matches. For example with printSomethingHelper being included in printSomething.

Llewellynvdm:
Please test the staging branch to see if the fix really resolved the issue for you.

PR #371, opened 2019-01-30
Converted all the JCB tables to have a DYNAMIC row format

Llewellynvdm:
Pull Request for Issue gh-365, gh-362, gh-360, gh-364, gh-367, gh-363, gh-369, gh-370.

### Summary of Changes
- Resolved gh-370 so that all fields get loaded in a dynamicGet, even new once created at a later stage.
- Added the option to set the row_format per table in the admin view.
- Converted all the JCB tables to have a DYNAMIC row format.
- Fixed gh-369 so that JCB packages like JMM can be imported.
- Resolved gh-363 so that show on will function as expected in the full width layout.
- Resolved gh-367 by adding a wrapper class around each field in the edit view.
- Resolved gh-364 to allow greater depth and ensure that all custom code placeholders are always update/replaced at any depth layer.
- Tweaked the editURL method to also handle access control permissions.
- Resolved gh-360 by adding the blob data type as an option for data types.
- Resolved gh-362 by adding the nested subform compatibility to subforms in JCB.
- Resolved gh-365 by improving the custom script validation search of save of custom code area. Added email validation to the email helper script.
- Fixed issue that caused some custom code not to export for a JCB packages if found in anther custom code area.
- Made some corrections to the router helper class.
- Added return_here value admin views to improve the return values in url redirection.
- Improved the open valid base64 method to make use of default string.
- Update many ajax calls to not use jsonp, and infact be more secure.

### Testing Instructions
Read related issues and test based on expected results explained in issues.

### Expected result
Check the related issues.

### Actual result
The all issues are resolved

I will be bumping a few new tutorials to explain some of the new features.


admin/helpers/compiler/a_Get.php
```diff
@@ -4936,19 +4977,13 @@ public function setCustomCodeData($string, $debug = 0, $not = null)
 	 * @return  string on success
 	 * 
 	 */
-	protected function insertCustomCode($string, $debug = 0)
+	protected function insertCustomCode($ids, $string, $debug = 0)
```
ro-ot: so simple!

Issue #410, opened 2019-04-26
Feature Request - auto setup database data in new field definition


jcodewalker:
Hi Llewellyn,

What do you think to add an auto fill the fields in the database tab, in field definition, based on the field type selected?

I mean, there are several field type that use often the same database setting, like radio: tinyint, length 1, default 0...
You could add a tab, in field type definition, to set default database value to be used to pre-set the database tab in field definition. There the user can always change the values.
What do you think?

Llewellynvdm:
Okay... give me the auto fill db values per field type, and I can see what would be the best path for implementation.

Llewellynvdm:
I have looked at this before, can't now remember why it did not happen. But I agree it would be nice to have :+1:

jcodewalker:
I'm not so expert and other can say different values, so I would say to use common sense.
What do you say about these?
type dbtype bdlength dbdefault
accesslevel: int 7 ''
calendar: datetime 50 ''
category: int 7 ''
checkbox: tinyint 1 0
checkboxes (prusume json encoded): text
color: varchar 7 ''
custom: int 11
customUser: int 11 0
editor : mediumtext
email: varchar 255 ''
file: varchar 255 ''
filelist: varchar 255 ''
folderlist: varchar 255 '' ?
hidden: int 1 0
integer: int 11 0
list: int 11 0
media: varchar 255 ''
menuitem: int 7 0
meter: int 11 0
number: int 11 0
password: varchar 255 ''
plugins: int 7 0
radio: tinyint 1 0
range: int 11 0
subform: mediumtext
tag: varchar 255 ''
tel: varchar 50 ''
text: varchar 255 ''
textarea: blob / text
timezone: int 7 0
url: varchar 255 ''
user: int 11 0
usergroup: int 11 0

Llewellynvdm:
aaah now I remember why we did not set these, we had an internal debate about the possible unexpected behavior if the user does not emphatically set the database values them selves. Since then they may expect that JCB will set the values based on the field attributes, which it does not do, and so we will be forced to build a new set of helper methods to based on field attributes tweak the database values. This will happen since the database is on the second tab, which they do not see when creating the field, so we are actually forcing the user to set the database every time since that seemed like the most trusted way to insure correct implementation. Therefor the note given under those fields about being sure to set the database values correctly.

We realized that this may not be a problem with all field types, but just with some, and yet if just one does not work well in this direction of automation, then we can't trust it fully.

[...]

Issue #407, opened 2019-04-12
Add PHP (getListQuery - JModelList) diff pos in dynamic get


jcodewalker:
Hi,

Add PHP (getListQuery - JModelList) in _**dynamic get**_ inserts the custom script in a different position respect to the same option in _**admin view / PHP**_.
While in admin view, with my custom script, I can access the whole query already set by JCB, in dynamic view I can't.

Sometimes we need to customize the query and, for example, in admin view I was able to refine the select('a.*') method adding a 'DISTINCT' mysql command, while in dynamic get I can't access the query as it is not jet populated by JCB. It's already set, but empty at this stage.

It should be possible to move the dynamic get injection to the same place of admin view?

Llewellynvdm:
Okay, I hear you. Let me say the dynamicGet was my first try at building a query builder, and wow it is still not fully there. So there are two things I need to achieve here, one I need to continue to expand the features of the dynamicGet area to be able to eventually do everything that is possible, or needed. The second I need to allow devs the freedom to rule the code.

That all said, I can add those areas you are asking for... but why not also add the needed, or lacking feature you want to custom code.

jcodewalker:
Well, I'm not able to say anything about other features to add as I'm still learning hot to use JCB but I think you already did an amazing work. Only, the dynamic get works a little bit different from the back end view. I suppose moving the injection from before the query building to after doesn't require a big effort and you can do it really quickly.
With this, you already achieve at least the freedom to rule the code and, I still suppose, the freedom to do everything (let's say manually). Great result, minimum effort.
Then, for the other features to add, there is Joomla v4 in the near feature... I would think about that in Jv4 mode, but really, you already did an amazing component. I have another hint but I'm going to open another thread.

Issue #404, opened 2019-04-05
Feature Request -- Dynamic Get -- Group


mwweb:
### Steps to reproduce the issue
I have a field in a table that doesn't have a separate joined table. In my instance, county. I want to create a list of items from only a single county.

I can create a dynamic get, but the list view will create a row for every instance of counties. For example, if there were 10 rows with county=County1, 5 rows with county=County2, the list view would show 15 rows, because that it what it correctly found.

If we had the addition of "Group By" in the Tweak tab, with only the table key, then I could enter a.county.

On compile, if group by has values, then a row similar to the following would be added along with where and order queries:

$query->group('a.county');

Currently I can add that code via custom code, but having a built in option would be cleaner.

### Expected result
Have the ability to group rows.

### Actual result
No option

### System information (as much as possible)
- OS Name & Version: 14.04.1-Ubuntu
- MySql Version:  5.5.5-10.1.34-MariaDB-1~trusty 
- Apache Version:  Apache/2.4.33 (Ubuntu) 
- PHP Version:  7.1.18-1+ubuntu14.04.1+deb.sury.org+1 
- Joomla Version: 3.9.11
- JCB Version: 2.9.8
- Browser: Firefox and Chrome

Llewellynvdm:
Why did you close this again?

mwweb:
I was debating whether it would be a worthwhile suggestion, or just something that would be rarely used.

mwweb:
I just reopened it. None of what I was trying was working, then today I finally got it working. So, I do think that this may be a good option. I'm not sure how many people would use it, but grouping items might be a good feature to have available.

Llewellynvdm:
Okay so the feature has been added, please test it from the staging branch, and let me know.

mwweb:
It looks like it did good:

`		$query->order('a.county ASC');
		$query->group('a.county');
`

Issue #394, opened 2019-03-08
Categories stored with unwanted target view appended to the extension name


pjdevries:
I added a Category Field to an Admin view, specifying the `extension` without a target view. The Admin view displays the category menu item, so I can add categories. However, the view, in which I use the Category Field, does not show the added categories.

On closer inspection, I see that the added categories are stored in the database _with_ the target view appended to the extension name, but I did not specify that in the Field definition. This is reflected  by the `<field>` definition in the generated xml file, which does not specifiy the target view either.

I attached some screenshots illustrating what I described.

### System information (as much as possible)
- OS Name & Version: JCB 2.9.11
- MySql Version: 5.7.21
- Apache Version: 2.4.33
- PHP Version: 7.2.4
- Joomla Version: 3.9.3
![Screenshot Editing the Field](https://user-images.githubusercontent.com/1171835/54055221-a9e5ae80-41ec-11e9-9e9f-cf56d4e0b26b.png)
![Screenshot phpMyadmin](https://user-images.githubusercontent.com/1171835/54055230-ae11cc00-41ec-11e9-9ece-14f4dede440d.jpg)
![Screenshot template xml](https://user-images.githubusercontent.com/1171835/54055233-b0742600-41ec-11e9-883c-c6c87f29c71b.jpg)

- JCB Version: JCB 2.9.13 **Beta 2019-03-06**
- Browser: FireFox

Llewellynvdm:
Please add the views name to the xml (field properties), that is how Joomla knows what categories you are looking for.

Far as I know your categories can only be per/view and not per/component. If this has changed please share the documentation... then I will see what should be done to fix this. But from what I remember categories are linked to views.

pjdevries:
It does work when appending the view name. But that is not my point. My point is that I think it is not necessary. Just the component name suffices. I am still trying to find concrete evidence, but here is why I think it is the case:

- The description of JCB's `category` field type itself. It says you _must_ enter the name of the extension and that you _can_ also target a view. Apparently you, or whoever wrote that description, also thought the view name appendix is optional.
- There is the description of Joomla's [Category form field type](https://docs.joomla.org/Category_form_field_type). There is no mention of the view name being mandatory for the `extension` attribute of the field.
[...]

PR #415, opened 2019-05-15
Many fixes, and improvements

Llewellynvdm:
Pull Request for Issue gh-404, gh-321, gh-407, gh-379, gh-410, gh-394.

### Summary of Changes
- Added Captcha field type. Started adding the crowdin integration.
- Fixed gh-394 so to be able to target the component instead of a single view.
- Resolved gh-410 to allow default data type values for the database to be set per fieldtype.
- Resolved gh-379 to allow null and zero values to pass via ajax call/post
- Fixed gh-407 by moving the getListQuery custom PHP position in the dynamicGet
- Updated the UIKIT versions
- Fixed the site view menu not being build as a result of new feature that builds edit site view menu.
- Resolved gh-321 by adding the option te set a create menu for admin edit views that are also added to the front-end.
- Resolved gh-404 by adding the option to use grouping in dynamicGet query.

### Testing Instructions
Please read over each issue and validate the expected outcome.


TLWebdesign:
excellent work :D

PR #483, opened 2019-10-11
Transliteration of cyrillic characters.

bamboo2panda:
Pull Request for Issue gh- .

### Summary of Changes
- Transliteration of cyrillic names of Views when preparing safe string

### Testing Instructions
- Make Admin view named in russian like "Таблица"
- Add this view to component
- Compile component
- Check folder admin/view

### Expected result
- Folders in admin/views:
    - tablica
    - tablici

### Actual result
- No folders of this view in admin/views

No changes required.


Llewellynvdm:
Please only commit the changes, meaning only the lines you changed or added, and do not reformat the whole file, since I can't see what you have changed.

Llewellynvdm:
I can add this no problem, we need also think of more areas that will be effected. So I am just wonder did you see the new implementation of [field naming](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/componentbuilder.php#L670) and [type naming](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/componentbuilder.php#L728) in JCB?

It looks like this is an older version of the helper class since it does no have these new function for field and type naming.

Always work on the latest branch of JCB staging :+1: 

I think we can wrap this option into a function to be turn-on in the GUI, since it has you have to anyway change the global language target right, in the global options of JCB:
image

So do you know which languages all use Cyrillic Characters the we can use that to trigger this feature.

Llewellynvdm:
Here is a list of all the Joomla languages in JCB:

af-ZA | Afrikaans
sq-AL | Albanian
ar-AA | Arabic Unitag
hy-AM | Armenian
id-ID | Bahasa Indonesia
eu-ES | Basque
be-BY | Belarusian
bn-BD | Bengali
bs-BA | Bosnian
bg-BG | Bulgarian
ca-ES | Catalan
zh-CN | Chinese Simplified
zh-TW | Chinese Traditional
hr-HR | Croatian
cs-CZ | Czech
da-DK | Danish
prs-AF | Dari Persian
nl-NL | Dutch
dz-BT | Dzongkha
en-AU | English AU
en-CA | English CA
en-GB | English GB
en-NZ | English NZ
en-US | English US
eo-XX | Esperanto
et-EE | Estonian
fi-FI | Finnish
nl-BE | Flemish
fr-FR | French
fr-CA | French CA
gl-ES | Galician
ka-GE | Georgian
de-AT | German AT
de-CH | German CH
de-DE | German DE
de-LI | German LI
de-LU | German LU
el-GR | Greek
he-IL | Hebrew
hi-IN | Hindi
hu-HU | Hungarian
ga-IE | Irish
it-IT | Italian
ja-JP | Japanese
km-KH | Khmer
ko-KR | Korean
lv-LV | Latvian
mk-MK | Macedonian
ms-MY | Malay
srp-ME | Montenegrin
nb-NO | Norwegian Bokmal
nn-NO | Norwegian Nynorsk
fa-IR | Persian
pl-PL | Polish
pt-PT | Portuguese
pt-BR | Portuguese Brazil
ro-RO | Romanian
ru-RU | Russian
sr-RS | Serbian Cyrillic
sr-YU | Serbian Latin
si-LK | Sinhala
sk-SK | Slovak
sl-SI | Slovenian
es-ES | Spanish
es-CO | Spanish CO
sw-KE | Swahili
sv-SE | Swedish
sy-IQ | Syriac
ta-IN | Tamil
th-TH | Thai
tr-TR | Turkish
tk-TM | Turkmen
uk-UA | Ukrainian
ug-CN | Uyghur
vi-VN | Vietnamese
cy-GB | Welsh
lt-LT | Lithuanian

We can add this also to the languages area, that we can turn this on and off in each language... just an idea. I could google this I suppose... but any help with this will be great :)

bamboo2panda:
This table is for Russian. The characters in other languages are almost the same, but they have a difference... Yes, it would be nice to make a more universal solution. 

But I think that it's no need in UI function.  Joomla has JLanguageTransliterate::utf8_latin_to_ascii that transform some strange characters to [a-zA-Z]. Like this here need a function that thansform any none a-z character to a-z or skip if can't do this.

So you do not need to turn it off, because this will lead to an error as in the description. 

Now I see that Joomla do the same thing in alias field!

Llewellynvdm:
Okay it seems like moving the whole feature to the GUI and making easy for any language to be set manually is idea... and will achieve the goal much more effectively.

I will push the changes in a few minutes.

Llewellynvdm:
> Now I see that Joomla do the same thing in alias field!

Okay great, let me take a look... it may save us some time.

Llewellynvdm:
Do you have a link to the code in Joomla?

bamboo2panda:
No, need to search...

Llewellynvdm:
This is how the GUI will look.

image

Just working on the implementation on the code side.

bamboo2panda:
Joomla makes alias by JFilterOutput::stringURLSafe($string)

I think it will be good just to process $string with this function. 
And... If it will be GUI feature, switch must be turn "Yes" by default. Because "No" may cause a component error. For example name "table_записи" will be compiled like "table_", that cause no error, but name "Записи" will be compilled to "" (empty string). 

P.s. sorry for my english =)

Llewellynvdm:
> Joomla makes alias by JFilterOutput::stringURLSafe($string)

I think this is the better solution for sure. I will drop the GUI idea... and instead use this option.

bamboo2panda:
Ok. Thanks!

Llewellynvdm:
@bamboo2panda thanks for the idea, and direction towards best implementation. Please try it out, remember you need to see the JCB global language to Russian for it to work.

From this:
image

To this:
image

Once you have tested it, let me know :+1:

bamboo2panda:
I tested this implementation on clear joomla with JCB. For testing I took Russian and German names of component and view. It works good! But it can be improved a little. 
When JCB compiles "transliterated" component, it make only localised language files (for example ru-RU....ini and ru_RU....sys.ini).
So users with over languages will see LANGUAGE_CONSTANTS insted of menu or fields names. It will be great if JCB make en-GB files with transliterated names as translation by default and whith real translations to en-GB when developer add some.
en-GB is default in Joomla, so if some component doesn't have another language files (ex. ru-RU), it will have English names and cause no GUI problems.

Llewellynvdm:
Hmmm I am not sure this is logic, for now JCB's behavior is true to the settings you make. If you say you are working in Russian, and so all your naming, and wording in an around is in Russian, it will correctly place them in the RU...ini file as it should. Then if you want it to also be in English, there is a translation area in JCB where you can export all the strings, and import it again with the English translation, then JCB will compile both the Russian and English .ini language files.

To change that will cause unexpected result, that is not logic.

We are in fact working on adding a bridge between Crowdin and JCB so that translation can be easily done via a community initiative. This require that we do not mix languages, and once we do we will lose control of these conventions. This will cause great confusion.

The only alternative for you is to add a Joomla plugin that change the JCB default behavior just for you, since we have events in the compiler that allow you to inject other behavior then the default as set forth by JCB in harmony with all.

I have not yet recorded any tutorial on how to do this, but it is straight forward. You have two events where you can force "English" ini file to be build instead.

- [jcb_ce_onBeforeBuildAllLangFiles](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler/f_Infusion.php#L1215)
- [jcb_ce_onBeforeSetLangFileData](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler.php#L114)

To target these events your plugin must be part of the extension group, active in Joomla and the added to the allowed JCB plugins in the global options of JCB.
image

This way you can hack the core, and yet not lose your changes with new updates. Hope this helps!

bamboo2panda:
Oh, I understand your translation strategy! It makes much more sense =) Thanks again!

Issue #571, opened 2020-06-14
errors with custom compiler folder path


jcodewalker:
### Steps to reproduce the issue
add a custom folder path to "Compiler folder path" field in JCB component configuration.

### Expected result
demo component correctly compiled .

### Actual result
long list of errors (obviously all files listed as missing are correctly in that folder as I moved all the administrator/components/com_componentbuilder/compiler folder content there following the instructions)
and
demo component half compiled (see attached images)
![empty-controllers-folder](https://user-images.githubusercontent.com/49115855/84587226-b314f080-ae1d-11ea-971f-e8a3193b6010.png)
![errors-list](https://user-images.githubusercontent.com/49115855/84587228-b4461d80-ae1d-11ea-8fea-033764e7488f.png)
![helpers-folder](https://user-images.githubusercontent.com/49115855/84587229-b5774a80-ae1d-11ea-9e1a-973e8aaacb26.png)
![layout-folder](https://user-images.githubusercontent.com/49115855/84587230-b60fe100-ae1d-11ea-9f32-8a209691ca94.png)
![models-folder](https://user-images.githubusercontent.com/49115855/84587231-b6a87780-ae1d-11ea-9bb2-e8244d2a52d3.png)
![views-folder](https://user-images.githubusercontent.com/49115855/84587234-b7d9a480-ae1d-11ea-8aa2-498ef9e12ada.png)

### System information (as much as possible)
- OS Name & Version: 
- MySql Version: 
- Apache Version: 
- PHP Version: 
- Joomla Version: both 3.9.19 and 3.9.11
- JCB Version: both 2.10.13 and 2.11.2
- Browser: 

### Additional comments
Compiling a more complex component with plugins and modules, I see that plugins and modules are correctly compiled, while the component is still half compiled like demo component.

Llewellynvdm:
Make sure that your web-server has access to that file location.

May I ask why you moved the compiler folder? When updating JCB the installer (of Joomla) will not move the new files (of the compiler) to update the remote folder, heads-up on that one.

Anyway are you on Mac or Windows? The file system is a little tricky to say the least, I would advice that if you are having issue just move stuff back to the default paths and all should be okay.

The when adding custom folders you will need to manually update those as you update JCB. Very important.

Then you must use full paths.

[...]

Issue #570, opened 2020-06-12
Module builder issue: postflight install wrong comment code


jcodewalker:
### Steps to reproduce the issue
add code in postflight install field

### Expected result
JCB comment code using php tags: /***jcbcomment***/

### Actual result
JCB comment added using html tags

Obviusly it's not possible to install it as is.

### System information (as much as possible)
- OS Name & Version: 
- MySql Version: 
- Apache Version: 
- PHP Version: 
- Joomla Version: 
- JCB Version: 2.10.13
- Browser:

jcodewalker:
Ok, it seems fixed in the latest release.

jcodewalker:
No, still there in version 2.11.2

Llewellynvdm:
Can you please give me a picture "screenshot" of this in the code and the JCB GUI.

Llewellynvdm:
No reply in 19 days so I assume issue is resolved, let me know if you have any further questions.

jcodewalker:
No reply to what?
You didn’t announced any update…

Llewellynvdm:
I asked the following question:

> Can you please give me a picture "screenshot" of this in the code and the JCB GUI.

Yet after 19 days I did not hear back from you with more details. So unless you give me enough details I can't reproduce the issue, so if I ask for more details, pictures, screenshots and actual code, then it means I can't fix this issue with the current information provided. So just from the start give me enough information, and we will have a better, and shorted path to resolving the issue.

jcodewalker:
![Schermata 2020-07-07 alle 07 51 28](https://user-images.githubusercontent.com/49115855/86722405-24b01b00-c027-11ea-9442-54b9b50c0c60.png)
As you can see I sent you the image 25 days ago.
Please, use a more steady system to manage issues.

Llewellynvdm:
@jcodewalker you are just so funny, you crack me up lol.

Github is used by Joomla, and will continue to be used for JCB. Maybe you should just go to [github](https://github.com/vdm-io/Joomla-Component-Builder/issues/570), like I do every time for you, and comment directly on the feed when images are ask for.

I will now look in to the issue, thank you for giving me the details.

Llewellynvdm:
Should be fixed now, please check and let me know.

jcodewalker:
Yes, fixed, confirm.
Thanks

Issue #569, opened 2020-06-11
Module Builder Issues: wrong field path for target admin


jcodewalker:
### Steps to reproduce the issue
Add custom fields to the module (module target: administrator area)

### Expected result
Fields and rules path
/administrator/components/com_mycomponent/models/fields
and
/administrator/components/com_mycomponent/models/rules
if "Fields and Rules path" is targeted to component
OR
/administrator/modules/mod_mymodule/fields
and
/administrator/modules/mod_mymodule/rules
if "Fields and Rules path" is targeted to module 

### Actual result
/administrator/components/com_mycomponent/modules/fields
and
/administrator/components/com_mycomponent/modules/rules
if "Fields and Rules path" is targeted to component
NOTE: modules instead of models

and
/modules/mod_mymodule/fields
and
/modules/mod_mymodule/rules
if "Fields and Rules path" is targeted to module 
NOTE: missing /administrator/

Result: the fields are not fired.

### System information (as much as possible)
- OS Name & Version: 
- MySql Version: 
- Apache Version: 
- PHP Version: 
- Joomla Version: on both 3.9.11 and 3.9.19
- JCB Version: on both 2.10.13 and 2.11.2
- Browser: 

### Additional comments
I was not able to find out where the path are set in compiler.
I need an hint on how to fix it  hardcoding the file(s) as I have to release a component to customer. Thanks.

jcodewalker:
Ok, I found where the bug is:
class getModuleMainXML
so, I fixed the file and now it works if I set the target path to component.
Didn't find where the module path is added though, it is still a bug to fix.

Llewellynvdm:
Can you give me a few screenshots about this. Meaning how you set it up on JCB, and then what JCB build, and then what you changed, and where you think the compiler should change. 

Thanks!

jcodewalker:
Hello,
for JCB, just add a module for administrator side to demo component. In module config, add custom fields to the form and point the fields path either to module or component, see attached image 8point-path-to-module) .

Compile.
The xml has a wrong absolute path, pointing to module in site side (xml-pointing-to-module-wrong)


The result is (result-with-xml-wrong):

the first 4 fields are custom fields and you can see they are not fired.


Now, to fix the bug, 
file: administrator/components/com_componentbuilder/helpers/compiler/e_interpretation.php
function: getModuleMainXML
line: 26400 (more or less, as I was testing other things when I shoot the screen…)

where the function set rule and field path to component:
[...]

Issue #565, opened 2020-05-28
'Class PHPExcel_IOFactory not found' error in Language Translations Import


maciejczub:
### Steps to reproduce the issue
Language Translations > Import Data

### Expected result
Import file with translated strings.

### Actual result
Error: Class 'PHPExcel_IOFactory' not found 

### System information (as much as possible)
- PHP Version: 7.3.14
- Joomla Version: 3.9.16
- JCB Version: 2.11.2
- Browser: Firefox 76.0.1

### Additional comments
Library dependency problem?

Llewellynvdm:
This is indeed a bug, to fix add the [following lines of code in the import_language_translations.php file](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/models/import_language_translations.php#L418), replace the whole method.
```php
	/**
	* Set the data from the spreadsheet to the database
	*
	* @param string  $package Paths to the uploaded package file
	*
	* @return  boolean false on failure
	*
	**/
	protected function setData($package,$table,$target_headers)
	{
		if (ComponentbuilderHelper::checkArray($target_headers))
		{
			// make sure the file is loaded
			ComponentbuilderHelper::composerAutoload('phpspreadsheet');
			$jinput = JFactory::getApplication()->input;
			foreach($target_headers as $header)
			{
				$data['target_headers'][$header] = $jinput->getString($header, null);
			}
			// set the data
			if(isset($package['dir']))
			{
				$inputFileType = IOFactory::identify($package['dir']);
				$excelReader = IOFactory::createReader($inputFileType);
				$excelReader->setReadDataOnly(true);
				$excelObj = $excelReader->load($package['dir']);
				$data['array'] = $excelObj->getActiveSheet()->toArray(null, true,true,true);
				$excelObj->disconnectWorksheets();
				unset($excelObj);
				return $this->save($data, $table);
			}
		}
		return false;
	}
```
Let me know if this resolves the issue.

maciejczub:
In the meantime I installed the PHPExcel library manually and added `jimport('phpexcel.library.PHPExcel');` in the code. This solved the problem but this patch above also works great. Thank you very much.

Llewellynvdm:
Yes we remove the old library just a few days ago, and this is still some of the growth pains. I would advice you to remove the old lib, as it is completely [abandoned by its developers](uhttps://github.com/PHPOffice/PHPExcelrl), and instead use the snipped I gave you :+1:

PR #581, opened 2020-07-28
Improved the export feature for all views, and much more...

Llewellynvdm:
Pull Request for Issue gh-576, gh-571, gh-570, gh-569, gh-565, gh-561.

### Summary of Changes
- Added the option to exclude language strings from JCB package exports.
- Added the security policy.
- Resolved gh-576 to allow empty filter options.
- Improved the export feature to allow bulk custom override, and dynamic behavior towards larger sets.
- Added function name to custom buttons as a class for CSS styling.
- Removed the fallback for empty description on title field.
- Improved the createUser method to allow custom method registerUser to create a user.
- Fixed the empty not_required field issue.
- Added option to dynamically set the sub-form layout from global options.
- Fixed bug in compiler that caused index warning.
- Fixed greateUser logic to use the admin and site creation modes in correct way.
- Fixed language import issue that cause header mismatching.
- Added the option to add header values to the custom-custom fields.
- Fixed gh-570 to add the correct custom field placeholder type, which is PHP and not HTML.
- Major improvement to user helper methods.
- Fixed custom field building options to allow options to be set.
- Fixed gh-569 to use correct paths in module custom fields and rules.
- Improved the fancy date helper methods.
- Added the new Text::_() language string search for language strings.
- Add the option to add header custom code to custom fields.
- Fixed gh-571 compiler path errors.
- Updated paypal donation link in readme note.
- Added notice to the compiler where it detects mismatching category view target names.
- Added the update script for categories in JCB.
- Made adjustments all across JCB to act according to gh-561 implementation.
- Fixed gh-565 by removing the old PHPExcel_IOFactory and adding the new.
- Updated the subform layout for various subforms in JCB for better display.
- Change the helper category naming conventions for better integration with Joomla fields. gh-561

### Testing Instructions
Read each issue addressed at test accordingly.

### Expected result
Many bug fixes, and improvements.


Llewellynvdm:
@ohrionmartin the batch issue, is it only happening in this staging branch? or in the existing stable version 2.11.2 of JCB as well?

Since if this issue exits in the stable version 2.11.2, then you should open a bug report, and we can try resolve it with the next update, so that we do not hold this new release/update back.

But if it is caused by this current new release/update, then we must first fix it before we can release version 2.11.3.

ohrionmartin:
@Llewellynvdm 
This seems to be a browser specific problem. I tested the current master branch, the stable staging branch (**v2.11.2**) and the beta release (staging branch **v2.11.3**) across **Firefox, Google Chrome and the Chromium web browser**. The **Batch Function** works perfectly on Chrome and Chromium, but malfunctioning on Firefox. So the problem seems to be with the browser and not **JCB**.

Issue #637, opened 2021-01-03
Edit site view (coresponding to an admin view) "save" and "save&close" buttons inactive


fred-the-coder:
### Steps to reproduce the issue
Add an admin wiew to a JCB component and select "Add+Menu" to the field "Edit/create Site View".

### Expected result
An edit site view is created in frontend (in components/_component name_/views/_admin edit view name_/tmpl/edit.php)

### Actual result
The edit site view is created, but it's "header" is "empty", leading to "save" and "save&close" buttons inactive.
The coresponding admin edit view "header" is:
```
JHtml::addIncludePath(JPATH_COMPONENT.'/helpers/html');
JHtml::_('behavior.tooltip');
JHtml::_('behavior.formvalidation');
JHtml::_('formbehavior.chosen', 'select');
JHtml::_('behavior.keepalive');
$componentParams = $this->params; // will be removed just use $this->params instead
?>
<script type="text/javascript">
	// waiting spinner
	var outerDiv = jQuery('body');
	jQuery('<div id="loading"></div>')
		.css("background", "rgba(255, 255, 255, .8) url('components/com_mosoftbluemarine/assets/images/import.gif') 50% 15% no-repeat")
		.css("top", outerDiv.position().top - jQuery(window).scrollTop())
		.css("left", outerDiv.position().left - jQuery(window).scrollLeft())
		.css("width", outerDiv.width())
		.css("height", outerDiv.height())
		.css("position", "fixed")
		.css("opacity", "0.80")
		.css("-ms-filter", "progid:DXImageTransform.Microsoft.Alpha(Opacity = 80)")
		.css("filter", "alpha(opacity = 80)")
		.css("display", "none")
		.appendTo(outerDiv);
	jQuery('#loading').show();
	// when page is ready remove and show
	jQuery(window).load(function() {
		jQuery('#mosoftbluemarine_loader').fadeIn('fast');
		jQuery('#loading').hide();
	});
</script>
<div id="mosoftbluemarine_loader" style="display: none;">
```

The coresponding front-end edit view "header" is :
```
defined('_JEXEC') or die('Restricted access');
###SITE_ADMIN__VIEW_HEADER###
?>
<div class="mosoftbluemarine-registre">
<?php echo $this->toolbar->render(); ?>
```

### System information (as much as possible)
- OS Name & Version: Win10
- MySql Version:  	5.6.19 
- Apache Version: Apache/2.4.10 (Win32) 
[...]

Issue #630, opened 2020-12-24
Using the templatestyle field type in JCB


Llewellynvdm:
I created [Templatestyle](https://docs.joomla.org/Special:MyLanguage/Templatestyle_form_field_type) field for adding in config of my component. but when add add a field with this type nothing show in config.
this is my field Properties:
![image](https://user-images.githubusercontent.com/27265318/102706282-b5ad4280-42a5-11eb-8d89-5fef43191720.png)
please help me
thanks

_Originally posted by @ArMaNKashef in https://github.com/vdm-io/Joomla-Component-Builder/discussions/623_

Llewellynvdm:
So it seems we had the field type in the wrong target group. 

> We therefore will move it from the [special array](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/componentbuilder.php#L3698) to the [plain array](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/componentbuilder.php#L3677), as the fix.

Issue #629, opened 2020-12-22
Permission settings in component view wont change by AJAX request


bM8qwa:
### Steps to reproduce the issue

- Compile the Demo component (or any other component)
- Create new Look or open an existing one
- Enter "Permissions" tab of the Look
- Select any usergroup and any action, try to change "Select new setting"

### Expected result
Selected setting appears in "Calculated setting" immediately without saving page.

### Actual result
Selected setting does not appear in "Calculated setting", spinner rolls. Though if you save the page, selected setting for permission saves too.

### System information (as much as possible)
- OS Name & Version: 
- MySql Version: 5.5.5-10.4.17-MariaDB 
- Apache Version: 2.4.46
- PHP Version: 7.4.13 
- Joomla Version: 3.9.23
- JCB Version: 2.12.3
- Browser: FF 84.0

### Additional comments
1. On selecting new setting, error appears in the console:
```
Uncaught TypeError: document.getElementById(...) is null 
in http://localhost/Site/media/system/js/permissions.js at 1:823
```
2. In global settings (in "Configuration") of the same component AJAX works like a charm, no problem.

Llewellynvdm:
This is a hateful issue since it is really a Joomla convention that assumes that all views that have permissions [have a title filed](https://github.com/joomla/joomla-cms/blob/3.9.23/media/system/js/permissions-uncompressed.js#L47). I have been asking multiple times to change things in J3. But due to the urgency to get J4 out, it is complicated.

I am looking at an option to resolve the issue... but honestly the workaround is a major issue. We have to first check if the view has a title field, and if it doesn't we need to add one, and in the getForm method remove the field again, to avoid database  conflicts... serious pain.

bM8qwa:
Thanks! Never could expect the permission behavior dependence on the view field set or the particular field name...

You certainly know, that the same wrong AJAX in permissions happens to JCB views themselves, let's say Joomla Components or Admin Views. 

Adding the "title" text field to the Look view in Demo component instantly fixes the permission things. Hopefully the field type of "title" could be "hidden" to use it as a patch, if I don't want the "title" in the view fields.

Closing this issue is up to you.

Llewellynvdm:
Ok so I added a fix... please test the staging branch an let me know. Painful workaround... but no other option at this stage.

bM8qwa:
Thank you.

[...]

Issue #626, opened 2020-12-21
Prevent translating date format while compiling


dormasta:
### Steps to reproduce the issue

Output some date in site view
echo HtmlHelper::date($DateString, Text::_('DATE_FORMAT_LC3'));
Compile

### Expected result
DATE_FORMAT_LC3

### Actual result
COM_COMPONENT_DATE_FORMAT_LCTHREE

### System information (as much as possible)
- OS Name & Version: 10.15.7
- MySql Version: 5.7.26
- Apache Version: 
- PHP Version: 7.4.2
- Joomla Version: 3.9.23
- JCB Version: 2.12.2
- Browser: safari, chrome, Firefox

Llewellynvdm:
Okays so the code that already is in JCB that should have work is found here on [line 6499](https://github.com/vdm-io/Joomla-Component-Builder/blob/staging/admin/helpers/compiler/a_Get.php#L6499) in the Get class.

Our problem is that [it removes numbers](https://github.com/vdm-io/Joomla-Component-Builder/blob/664c2d5f2a92171d32c6c84d7dfbba9fba284069/admin/helpers/componentbuilder.php#L7436) and so since `DATE_FORMAT_LC3` has a 3 in it, it gets converted to `DATE_FORMAT_LCTHREE` and therefore fails the test...

Can you try the following as a fix?
```php
public function setLang($string)
{
	// this is there to insure we don't break already added Language strings
	if (ComponentbuilderHelper::safeString($string, 'U', '_', false) === $string)
	{
		return false;
	}
	// build lang key
	$keyLang = $this->langPrefix . '_' . ComponentbuilderHelper::safeString(
			$string, 'U'
		);
	// set the language string
	$this->setLangContent($this->lang, $keyLang, $string);

	return $keyLang;
}

```
Let me know if this works for you... so I can update the core.

dormasta:
HI Llewellyn,

absolutely sorry for the delay, but I somehow crashed my component and it took me some time to find the reason.
Anyway, you fix works like a charm. Thank you very much.

Cheers,
DTA

PR #641, opened 2021-01-03
Bug fixes

Llewellynvdm:
Pull Request for Issue gh-629, gh-630, gh-626, gh-637 .

### Summary of Changes
- Fixed gh-637 so that the header values will be added.
- Updated the copyright to only show the first year so we need not update these again.
- Improved the layout compiler methods.
- Added the trash helper option to JCB as a pro feature.
- Added more header options and renamed the header method in the compiler.
- Improved the compiler message area.
- Fixed gh-626 to ensure that language strings that are already uppercase do not get added, even if it has a number.
- Fixed gh-630 by moving the templatestyle field type to the plain field group.
- Improved the permissions adding behaviour so that no permissions tab will be added if no permissions are set in a view. gh-629
- Fixed gh-629 to add a hidden title field when a view does not have a title field.
- Updated the forum details

### Testing Instructions
Please read the related issues, and test accordingly. 

### Expected result
That all issues are fixed


ohrionmartin (approved):
`###SITE_ADMIN_VIEW_HEADER###` placeholder now gets replaced by the appropriate header code and all the toolbar buttons are now active and functional. The trash helper methods also work very well, thanks!

Issue #735, opened 2021-04-26
Typo in notice in compiler


TLWebdesign:
This line has typo "folder/s" instead of folders. 
https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/helpers/compiler/b_Structure.php#L2102 

This line also had the "folder/s" typo and component is not spelled correctly towards the end of the sentence. 
https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/helpers/compiler/b_Structure.php#L2107

typo "component"
https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/views/compiler/view.html.php#L187
https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/views/compiler/view.html.php#L189
https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/site/helpers/componentbuilder.php#L1116
https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/helpers/compiler/c_Fields.php#L6138

Llewellynvdm:
> This line has typo "folder/s" instead of folders.
> 
> https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/helpers/compiler/b_Structure.php#L2102
> 
> This line also had the "folder/s" typo and component is not spelled correctly towards the end of the sentence.
> 
> https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/helpers/compiler/b_Structure.php#L2107

The folder/s means one folder or many folders... since we are not counting them. So this one is not a typo.


> typo "component"
> 
> https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/views/compiler/view.html.php#L187
> 
> https://github.com/vdm-io/Joomla-Component-Builder/blob/48d92659838a993ea98a9e6691b0209c0d381aeb/admin/views/compiler/view.html.php#L189
> 
[...]

PR #741, opened 2021-05-04
Typo fixes (Ref #735)

TLWebdesign:
Pull Request for Issue gh-735 .

### Summary of Changes

Typo's corrected.

### DIFF
--- a/admin/helpers/compiler/b_Structure.php
+++ b/admin/helpers/compiler/b_Structure.php
@@ -2099,12 +2099,12 @@ class Structure extends Get
 							// set message that this was done (will still add a tutorial link later)
 							$this->app->enqueueMessage(
 								JText::_(
-									'<hr /><h3>Dynamic folder/s were detected.</h3>'
+									'<hr /><h3>Dynamic folders were detected.</h3>'
 								), 'Notice'
 							);
 							$this->app->enqueueMessage(
 								JText::sprintf(
-									'A method (setDynamicF0ld3rs) was added to the install <b>script.php</b> of this package to insure that the folder/s are copied into the correct place when this componet is installed!'
+									'A method (setDynamicF0ld3rs) was added to the install <b>script.php</b> of this package to insure that the folders are copied into the correct place when this component is installed!'
 								), 'Notice'
 							);
 						}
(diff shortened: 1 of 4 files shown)


TLWebdesign:
I hope i followed correct procedures. My first pull request in my life. I tried linking an issue but can't seem to do so.

admin/helpers/compiler/b_Structure.php
```diff
@@ -2099,12 +2099,12 @@ private function setStatic()
 							// set message that this was done (will still add a tutorial link later)
 							$this->app->enqueueMessage(
 								JText::_(
-									'<hr /><h3>Dynamic folder/s were detected.</h3>'
+									'<hr /><h3>Dynamic folders were detected.</h3>'
```
Llewellynvdm, line 2102: This is not a typo... the intend is that is is both folder and folders so the / is an or. So you can remove this change.

admin/helpers/compiler/b_Structure.php
```diff
@@ -2099,12 +2099,12 @@ private function setStatic()
...
 							$this->app->enqueueMessage(
 								JText::_(
-									'<hr /><h3>Dynamic folder/s were detected.</h3>'
+									'<hr /><h3>Dynamic folders were detected.</h3>'
 								), 'Notice'
 							);
 							$this->app->enqueueMessage(
 								JText::sprintf(
-									'A method (setDynamicF0ld3rs) was added to the install <b>script.php</b> of this package to insure that the folder/s are copied into the correct place when this componet is installed!'
+									'A method (setDynamicF0ld3rs) was added to the install <b>script.php</b> of this package to insure that the folders are copied into the correct place when this component is installed!'
```
Llewellynvdm, line 2107: This is not a typo... the intend is that is is both **folder** and **folders** so the **/** is an **or**. So you can remove this change to **folder/s**.

Llewellynvdm:
The rest looks good :+1:

admin/helpers/compiler/b_Structure.php
TLWebdesign: It's probably better to reply here :) You're right. I missed that. After some research i did find that the forward slash should be read as an "or". As in folder/folders. From the research i did online it is better to use "(s)" instead of "/s" as the parentheses means it's "optional" as in it could be both with or without s. instead of "one or the other". Thats why i propose to change it to (s).

Llewellynvdm (approved):

PR #1010, opened 2025-12-22
Fix: Add use GuidHelper in compiler

Scuottolinx:
Issue: gh-1277

When creating a subform and linking a structured field inside it (for example, an SQL field or other structured/dynamic fields), the Joomla Component Builder compiler throws a fatal error and stops the compilation process.

The expected behavior is that the compiler correctly handles structured fields inside subforms and completes the component compilation without errors.

Compiler error returned:

Class "VDM\Joomla\Componentbuilder\Compiler\Creator\GuidHelper" not found 

This error suggests that a required dependency is missing or not properly loaded during the subform field normalization process.

Steps to reproduce the Bug
Create a subform with a fieldCreate a new Subform field in Joomla Component Builder.
Inside the subform, add a structured field (e.g. an SQL field or another structured/dynamic field).
Save the field and include it in a fieldset or view of the component.
Start the component compilation process.
The compiler stops and returns the following error:
Class "VDM\Joomla\Componentbuilder\Compiler\Creator\GuidHelper" not found
The compilation fails and the component is not generated.


Llewellynvdm:
This repo is a mirror, all pull requests must please be made against https://git.vdm.dev/joomla/Component-Builder

Llewellynvdm:
Thank you for this!

